PR #3, opened 2021-02-08
import: sssd internals

pbrezina:
This is a very long document, if I recall correctly it was written mostly
for some certification (the internals needed to be documented) but honestly,
it is not well orginized for a new comer.

At this point I fixed formatting where possible but left the content intact,
therefore you do not need to proof-read it, just check the formatting.

I will get back to this when we start creating new content to split it into
multiple smaller and organized documents, possible with a more educating
content.


elkoniu (commented):
LGTM

PR #6, opened 2021-02-17
Add IPC document

pbrezina:
I rewrote this document quite a lot. It contained lots of low level
information that was too much code related so I removed those parts.

I also changed the wording a lot to something that imho provide more
and better information.


src/contrib/ipc.rst
```diff
@@ -0,0 +1,153 @@
...
+    server, the data is always written to the cache by the backend process and
+    returned to the calling library in the responder process.
+
+Helpers
+    SSSD runs in asynchronous mode so it is able to serve multiple requests in
+    parallel. However, this is not achieved by threads but by event driven
+    programming. Therefore operations that cannot be executed in a non-blocking
+    way are run in a special helper process. The helper process is spawned so it
+    can perform an otherwise blocking operation (e.g. ``kinit``). Example of such
+    process are `krb5_child` and ``ldap_child``.
```
justin-stephenson, line 44: The formatting of krb5_child and ldap_child differs here

src/contrib/ipc.rst
```diff
@@ -0,0 +1,153 @@
...
+All SSSD components listen to several standard signals. It is usually enough
+to send the signal to the main monitor process, which will then propagate it
+to other components.
+
+SIGTERM
+    Terminates a process gracefully.
+
+SIGKILL
+    In cases where an unresponsive worker process does not terminate
+    after receiving SIGTERM, the monitor forcibly kills it with SIGILL.
```
justin-stephenson, line 138: s/SIGILL/SIGKILL

justin-stephenson:
Just the 2 comments above, document looks great otherwise.

justin-stephenson:
One last comment, I'm not sure if the table of contents at the top of the page with the links to each section is very useful here. It is a small document and I doubt people would need to jump around. I'll defer to your decision on this however.

pbrezina:
Thank you. Fixed.

About ToC - I removed it for now. We should revisit the local toc idea and format and automate it when there is more content.

justin-stephenson:
Ack from my side.

PR #12, opened 2021-03-05
Add contribution guideline

elkoniu:
This is migration of old documentation including:
https://sssd.io/docs/developers/contribute.html
https://sssd.io/docs/newcomers/getting_started.html#github-workflow

### DIFF
--- a/src/contrib/contribution-policy.rst
+++ b/src/contrib/contribution-policy.rst
@@ -1,0 +1,42 @@
+Contribution policy
+===================
+
+There are many ways you can contribute to the System Security Services Daemon.
+
+SSSD development discussions occur either on:
+
+- SSSD GitHub project `issues <https://github.com/sssd/sssd/issues>`_
+
+- SSSD development mailing list
+  (`sssd-devel <https://lists.fedorahosted.org/archives/list/sssd-devel@lists.fedorahosted.org/>`_)
+
+- #sssd `IRC channel <irc://irc.freenode.net/sssd>`_ on
+  `freenode <http://freenode.net/>`_.
+
+Keep in mind that SSSD developers are spread across different time zones.
+Mailing list is a preferred communication channel if you wish to reach the entire team.
+
+The SSSD originated as a sister-project to `FreeIPA <http://www.freeipa.org/>`_,
+so we share many pieces. Contributing to FreeIPA may also help SSSD and vice versa.
+
+If you want to file a bug or enhancement request please use SSSD GitHub
+`issue <https://github.com/sssd/sssd/issues>`_ tracking system.
+
+Licensing
+---------
+
+All source code committed to the SSSD is assumed to be made available under
+the GPLv3+ license unless the submitter specifies another license at that time.
+The upstream SSSD maintainers reserve the right to refuse a submission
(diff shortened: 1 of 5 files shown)


elkoniu:
I am not sure if bug report is in the scope of migration as it refers mostly "before GitHub era":
https://sssd.io/docs/users/reporting_bugs.html
If it should be migrated I think it should be part of contribution guideline and I can include it in this PR.
@pbrezina what do you think?

elkoniu:
One thing I do not included in this PR is "Notes for Maintainers" section from:
https://sssd.io/docs/newcomers/getting_started.html#github-workflow
I am not sure if this should be part of contributor-facing documentation, please advice here.

src/contrib/contribution-policy.rst
```diff
@@ -0,0 +1,43 @@
+Contribution policy
+===================
+
+There are many ways you can contribute to the System Security Services Daemon.
+This page should provide some basic pointers.
```
justin-stephenson, line 5: I think this line can be removed, s/This page should provide some basic pointers//

justin-stephenson:
Besides the `Tasks for newcomers` section, the contribution-policy document does not actually contain any information about contributing. Maybe the github-workflow content should be merged into this doc.

justin-stephenson:
Regarding the github-workflow content, I wonder if we really need to provide all of the github commands involved in the github fork & pull request workflow, IMO we could clean up the document if we only include the noteworthy SSSD specific steps/information, for example we would keep:

`Before submitting a patch, always make sure it doesn’t break SSSD tests and applies to the latest upstream master branch.`

But we don't really need `Setting up git` or `Downloading the code` sections, that information can be found in several other places with arguably better github fork + PR workflow information.

src/contrib/contribution-policy.rst
elkoniu: Fixed

elkoniu:
@justin-stephenson Small explanation from my side is needed I think.

Scope of this PR was to migrate current contribute documentation [1].
I found GitHub workflow document [2] which overlaps with [1] so I merged it in. At the end I split combined [1] and [2] content into few files to make it more clean and easier to update later.

Current state is that nothing is missing from old documentation - just the form is different.
Maybe my way of splitting this into multiple files was wrong - I do not know.
I agree with you that it needs to be refactored. For example we should assume that developer is familiar with git / GitHub. If not - the same way we should provide C programming tutorial etc.
I think documentation refining will be the next step after migration will be completed.

[1] https://sssd.io/docs/developers/contribute.html
[2] https://sssd.io/docs/newcomers/getting_started.html#github-workflow

pbrezina:
> I agree with you that contribute guideline needs to be refactored. For example we should assume that developer is familiar with git / GitHub. If not - the same way we should provide C programming tutorial etc.

Don't assume, but provide link to existing github documentation instead of providing yet another one.

> I think documentation refining will be the next step after migration will be completed.

Yes. But if there's already something that you'd like to do, feel free to do it.

pbrezina:
I talked to Pavel and he has no plan to further update this document. Ack.

We will revisit the document later when we finish migration of old  documents.

PR #7, opened 2021-02-22
Add Reporting Bugs page

justin-stephenson:
* Replaced FAS/Fedora account mentions with GitHub.
* Combined some sections and removed extra unnecessary wording
* Minor grammatical changes

PR #16, opened 2021-03-12
[WIP] Add bug reporting guideline

elkoniu:
This is migration of old bug reporting procedure description:
https://sssd.io/docs/users/reporting_bugs.html

It is right now out of the scope of migration but after small refactoring it may be useful
to have this as part of new documentation.


pbrezina:
We already have this merged: https://github.com/SSSD/sssd.io/pull/7

elkoniu:
True, my mistake as I was on old branch without this change already merged:)

PR #21, opened 2021-03-23
Add cache req document

justin-stephenson:
I wrote this based on my understanding of reading the code, please let me know if I missed any important components and I will add it. For example I did not give a detailed explanation about domain selection/location codepaths.

### DIFF
--- a/src/contrib/cache_req.rst
+++ b/src/contrib/cache_req.rst
@@ -1,0 +1,396 @@
+.. contents:: Table of Contents
+    :local:
+
+Unified Cache Interface
+#######################
+
+The unified cache interface (``cache_req``) is an integral component of the SSSD data flow, consumed by most SSSD responders including NSS, PAM, InfoPipe, autofs and ssh.
+
+The ``cache req`` component implements a generic interface for SSSD requests. The interface is responsible for handling the logic for domain selection, cache lookups, and sending requests to the backend. ``cache_req`` implements a plugin architecture where a plugin exists for each object request type. The plugin dictates the behavior of cache request, which attributes should be used to lookup the object in the cache, etc. Each request follows similar logic through the cache request interface, making it easier to understand and debug.
+
+.. mermaid::
+
+    graph LR
+        app(Application)
+        gnss(glibc-NSS)
+        snss(sssd_nss)
+        cache_req(cache_req)
+        lcache(local cache)
+        be(backend)
+
+        style gnss fill:green,stroke:green,stroke-width:1px,fill-opacity:0.2
+        style snss fill:green,stroke:green,stroke-width:1px,fill-opacity:0.2
+        style cache_req fill:blue,stroke:blue,stroke-width:1px,fill-opacity:0.2
+        style lcache fill:yellow,stroke:yellow,stroke-width:1px,fill-opacity:0.2
+        style be fill:purple,stroke:purple,stroke-width:1px,fill-opacity:0.2
+        
+        app --> gnss
+        gnss --> snss
+        snss --> cache_req
+        cache_req --> lcache
+        lcache <--> be
+        be --> lcache
+
+
+How to use the cache_req API
+****************************
+
+Initiating a cache request is done by calling the tevent request input function ``cache_req_send``
+
+.. code-block:: c
+
+    struct tevent_req *cache_req_send(TALLOC_CTX *mem_ctx,
+                                      struct tevent_context *ev,
+                                      struct resp_ctx *rctx,
+                                      struct sss_nc_ctx *ncache,
+                                      int midpoint,
+                                      enum cache_req_dom_type req_dom_type,
+                                      const char *domain,
+                                      struct cache_req_data *data);
+
+Arguments
+=========
+
+midpoint
+    Handle background (out of band) object refreshes. It is used to enable the ``entry_cache_nowait_percentage`` configuration option in the NSS responder. Set this argument to 0 for non-NSS responder callers.
+
+req_dom_type
+    Limit the request type lookups to only POSIX Domains (``CACHE_REQ_POSIX_DOM``), only ``CACHE_REQ_APPLICATION_DOM`` or any domain type (``CACHE_REQ_ANY_DOM``). POSIX domains are reachable by all services. Application domains are only reachable from the InfoPipe responder and the PAM responder. Only objects from POSIX domains are available to the operating system interfaces (NSS) and utilities.
+
+domain
+    Request searching an explicit domain, NULL otherwise. When ``domain`` is NULL we delegate the logic of domain selection to ``cache req``.
+
+data
+    Must be a ``struct cache_req_data`` formatted data. Several helper functions exist to create this cache_req_data (See ``cache_req_data_*`` in `cache_req.h <https://github.com/SSSD/sssd/blob/master/src/responder/common/cache_req/cache_req.h>`_) based on the type of input data.
+
+Retrieving the output from the cache request is done with either of the two ``_recv`` functions:
+
+.. code-block:: c
+
+    errno_t cache_req_recv(TALLOC_CTX *mem_ctx,
+                           struct tevent_req *req,
+                           struct cache_req_result ***_results);
+
+    errno_t cache_req_single_domain_recv(TALLOC_CTX *mem_ctx,
+                           struct tevent_req *req,
+                           struct cache_req_result **_result);
+
+You may want to only skip searching the local cache, or the data provider. The public functions in ``cache_req.h`` can be used to tune the behavior of the request lookups. See ``lookup_behavior``
+
+.. code-block:: c
+
+    void
+    cache_req_data_set_bypass_cache(struct cache_req_data *data,
+                                    bool bypass_cache)
+
+    void
+    cache_req_data_set_bypass_dp(struct cache_req_data *data,
+                                 bool bypass_dp)
+
+    void
+    cache_req_data_set_requested_domains(struct cache_req_data *data,
+                                         char **requested_domains)
+
+A simple caller example (user initgroups request) may look as follows:
+
+.. code-block:: c
+
+    /* Prepare the input data */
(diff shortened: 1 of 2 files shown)


pbrezina:
I think this is a good start but imho the focus is different then what it should be.

The target reader is a developer (external and internal) who wants to contribute to the project and use cache_req in his code or even implement a cache_req plugin. So I think you don't need to focus on the NSS responder at all, but put all energy into describing the cache_req in detail - not only the basic search but all the details that happens under the hood, all the plugins capabilities and switches and so on.

The goal is to really describe how it works under the hood - even I don't understand it completely since it changed a lot from the time I wrote it :-) And ideally, make it easily maintenable.

BTW I described the basic lookup in SSSD architecture, you can just link it from this document when its merged and focus on details here.

pbrezina:
I think this document should consist from two parts:

1. How to use cache_req API
2. What happens under the hood and how to write new plugins and extend functionality

justin-stephenson:
Thank you, I'll re-work the document and submit an update once done.

pbrezina:
> > I think this document should consist from two parts:
> > ```
> > 1. How to use cache_req API
> > ```
> 
> From internally within SSSD is acceptable? Or is this intended to be an external API for any application to call?

Internal only.

> > What happens under the hood and how to write new plugins and extend functionality
> 
> I thought that a cache_req plugin exists to handle each type of NSS library call, what would be the use case for writing a new plugin? Is `CACHE_REQ_USER_BY_CERT` an example of this type of use case?

It's not only about nss, it handles everything (expect sudo for now). So if something new is added, it needs to get support in cache req.

pbrezina:
Perhaps, we can split it into two documents:
* how to use cache req API into "fundamentals" section
* how does it work internally into "under the hood" section

src/contrib/cache_req.rst
```diff
@@ -0,0 +1,290 @@
...
+
+The ``cache req`` (request) interface provides an internal API to request identity
+information from SSSD, either directly from the local SSSD cache or retrieving refreshed data
+from the data provider. The SSSD NSS responder is the primary consumer of the interface,
+however ``cache_req`` also refreshes information which may be needed by the PAM
+responder, or provide user information attributes over the SSSD infopipe.
+
+.. mermaid::
+
+	graph LR
```
elkoniu, line 15: This chart looks "too flat" to me. I think similar to the one [1] made by jhrozek in the past would be more meaningful.
[1] https://jhrozek.files.wordpress.com/2015/03/sssd-lookup.png

src/contrib/cache_req.rst
```diff
@@ -0,0 +1,290 @@
...
+
+NSS responder request setup
+---------------------------
+
+In preparation for setting up a request, The NSS responder calls into the function
+``nss_cmd_getpwnam()``. This function is a wrapper of the generic function
+``nss_getby_name()`` with the  argument ``cache_req_type`` enum ``CACHE_REQ_USER_BY_NAME``.
+The cache request types can be understood as a mapping from the equivalent NSS
+library interface lookup functions. ``getgetnam()`` sets ``CACHE_REQ_GROUP_BY_NAME``,
+``nss_cmd_getgrgid(``) sets ``CACHE_REQ_GROUP_BY_ID()``, etc...
```
elkoniu, line 36: I think I would put here full definition of `enum cache_req_type` in some scrollable box or table with some mappings in between functions and `cache_req` types.

src/contrib/cache_req.rst
```diff
@@ -0,0 +1,290 @@
...
+to begin the cache request operation.
+
+Sending a cache request
+-----------------------
+
+``cache_req_send()`` is the top-level function to initiate a cache request, in our
+example it is called by the NSS responder ``nss_getby_name()`` function. The
+**simplified** ``cache_req_send()`` logic is described below:
+
+.. code-block:: sssd-log
```
elkoniu, line 92: This code snippet not fits the description above.

src/contrib/cache_req.rst
```diff
@@ -0,0 +1,290 @@
...
+
+	[cache_req_select_domains] (0x0400): CR #10: Performing a multi-domain search
+
+.. note::
+    SSSD lookups may bypass the cache or the data provider. This determination is based on the ``cache_behavior`` enum which is also set at this time.
+    This can be seen in the logs here, `check` may be replaced with `bypass`.
+
+	*[cache_req_search_domains] (0x0400): CR #23: Search will check the cache and check the data provider*
```
elkoniu, line 189: This should be `sssd-log` snippet I think.

src/contrib/cache_req.rst
```diff
@@ -0,0 +1,290 @@
...
+
+    [cache_req_set_domain] (0x0400): CR #0: Using domain [implicit_files]
+
+Performing the search
+---------------------
+
+In a multi-domain search, the below logic repeats for each domain until a result is found, or SSSD searched all
+domains and no result was found. Let's assume the normal cache behavior(do not bypass).
+
+.. mermaid::
```
elkoniu, line 204: I would use a different graph skin for the control flow and a different one for the data flow.

src/contrib/cache_req.rst
```diff
@@ -0,0 +1,290 @@
...
+    B --Not found--> D{Data Provider}
+    B --Object found--> C[Expired?]
+    C --Object expired--> D
+    D --Object found in dp--> E[Update local cache]
+    D --Object not found --> F[Next domain]
+    F --> A
+    E --> G[Search local cache again]
+
+
+* First the search checks the negative cache(``cache_req_user_by_name_ncache_check()``)
```
elkoniu, line 218: Graph mixed with bullet list and log snippets make this chapter content a bit hard to follow.

justin-stephenson:
I did a large rewrite of the document, I believe it is more aligned with the expected content now.

I would appreciate help in completing writing about the ``Well known object`` code and proper use of the cache_req plugin well_known_fn - I did not really understand what this code does, or what problem is solves.

Thank you for the review @elkoniu but I suspect your review items may not be relevant anymore as I removed most of the previous content.

src/contrib/cache_req.rst
```diff
@@ -0,0 +1,382 @@
...
+
+Preparing the domain
+--------------------
+
+Before performing the object search inside the domain, the ``prepare_domain_data_fn`` is called, this is used by some plugins to to alter lookup data per specific domain rules, such as case sensitivity, fully qualified format etc.
+
+Domain locator
+--------------
+
+The domain locator plugin exists to alleviate performance problems when SSSD must iterate over several domains(ex: AD forest with multiple domains) for each unqualified name, or by-ID lookups. The domain locator can help to find the correct domain to search early on in the lookup flow, instead of iterating through all domains.
```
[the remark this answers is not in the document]
pbrezina: each unqualified name, or by-ID lookups -> each unqualified name or by-ID lookups

src/contrib/cache_req.rst
```diff
@@ -0,0 +1,382 @@
...
+
+    const struct cache_req_plugin cache_req_object_by_id = {
+        ...
+        .dp_get_domain_check_fn = cache_req_object_by_id_get_domain_check,
+        .dp_get_domain_send_fn = cache_req_object_by_id_get_domain_send,
+        .dp_get_domain_recv_fn = cache_req_common_get_acct_domain_recv,
+
+A cache-only search is performed first to check if the object already exists in the cache. If not, then we execute the domain locator plugin ``dp_get_domain_check_fn`` function to check if the id exists in the negative cache. If it is not in the negative cache, then we call the ``dp_get_domain_send_fn``
+
+The send function sends a SBUS ``getAccountDomain`` request to the backend, if a domain is reported as containing an object, all domains except that
```
pbrezina, line 259: a SBUS -> an SBUS

src/contrib/cache_req.rst
```diff
@@ -0,0 +1,382 @@
...
+one are marked with negative cache entries for that request(using the associated ``.ncache_add_fn`` plugin function``.
```
pbrezina, line 260: plugin function``. -> plugin function.

src/contrib/cache_req.rst
```diff
@@ -0,0 +1,382 @@
...
+
+The domain locator plugin code is only executed for unqualified requests with multiple domains, or on the second pass of a ``CACHE_REQ_CACHE_FIRST`` lookup.
```
pbrezina, line 262: domains, or on the -> domains or on the

src/contrib/cache_req.rst
```diff
@@ -0,0 +1,382 @@
...
+
+* First the search checks the negative cache using the plugin ``ncache_check_fn`` function. If the object exists in the negative cache, and the negative cache timeout has not been reached then cache_req will return ``ENOENT``.
+
+.. code-block:: c
+
+    .ncache_check_fn = cache_req_user_by_name_ncache_check,
+
+* Next, search the cache to see if the object already exists in the local cache. Here, the plugin-defined ``lookup_fn`` is used to handle the different object types and ``SYSDB`` attributes to search. If the object is found and is not expired, it can be returned successfully.
+
+.. warning::
```
pbrezina, line 296: Place this before or after the note.

pbrezina (changes requested):
cache_req -> \``cache_req``

Minor comments in review. You did a good job there.

src/contrib/cache_req.rst
```diff
@@ -0,0 +1,382 @@
...
+If ``allow_missing_fqn`` is set to false, then this multi domain search iterating through domains will skip domains which require fully qualified names.
+
+``get_next_domain_flags`` sets the flags which are passed to ``get_next_domain()`` during iteration. At the time of this writing, cache request plugins set this to ``0``, or ``SSS_GND_DESCEND`` to determine behavior when iterating through multiple domains.
+
+.. table::
+    :align: left
+    :widths: 1, 3
+    :width: 70%
+
+    =========================== ====================================
```
pbrezina, line 232: Add a header. It will look better.

justin-stephenson:
Thanks very much for the review, review items addressed as requested.

pbrezina:
Ack.

PR #39, opened 2021-07-16
[WIP] docs: Using shortnames in SSSD

elkoniu:
This document adds description about using shortnames in SSSD.

### DIFF
--- a/src/troubleshooting/shortnames.rst
+++ b/src/troubleshooting/shortnames.rst
@@ -1,0 +1,118 @@
+SSSD and shortnames
+###################
+
+In the simplest configuration SSSD can serve single domain.
+In this case user / group query will be unambiguous as only one domain can provide the answer.
+Multi domain setup is more complicated as the same user / group can be registrated in multiple domains.
+To make query unique in this scenario fully qualified name need to be used:
+
+.. note::
+
+    Fully "qualified" user name is the situation when username is followed with domain name,
+    for example bob@corp.abc.com instead of just "bob". On Windows systems it can be called
+    as SAM account (Security Account Manager) and writted different way: domain_name\user_name.
+    UPN (User Principal Name) is another alias for this naming convention.
+    In case of Active Directory (AD) this type of entry may be called DN
+    (Distinguished Name) where separated blocks will have value:
+    CN=bob,OU=Users,DC=corp,DC=abc,DC=net
+
+When fully qualified user name is used query will be send only to single domain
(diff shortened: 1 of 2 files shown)


elkoniu:
I am still updating this document. At the end it will contain few user scenario descriptions and more details about shortnames when backend provider is IPA.

elkoniu:
@pbrezina Alpha is ready, please take a look if the general structure looks OK to you.

src/troubleshooting/shortnames.rst
```diff
@@ -0,0 +1,287 @@
...
+
+.. note::
+
+    Fully qualified user name is the situation when user name is followed with domain name,
+    for example ```bob@corp.abc.com`` instead of just ``bob``. On Windows systems it can be called
+    ``SAM`` account (``Security Account Manager``) and writted different way: ``domain_name\user_name``.
+    UPN (User Principal Name) is another alias for this naming convention.
+    In case of Active Directory (AD) this type of entry may be called ``DN``
+    (``Distinguished Name``), where separated blocks will have value:
+    CN=bob,OU=Users,DC=corp,DC=abc,DC=net
```
pbrezina, line 14: Just mention the two formats: bob@domain and DOMAIN\bob (in the second case it is not really a domain name but a netbios name, which is little bit different thing - also called a flatname in sssd code base).

SAM and UPN are not relevant for description of a fully qualified name. SAMAccountName is attribute that holds username in AD. UPN is attribute that stores Kerberos principal name.

pbrezina (changes requested):
I think you focused on different and much simpler topic - i.e. using `default_domain_suffix` and it is not very informative as such. Also you certainly don't want to touch nsswitch.conf for reasons that you described. 

What this article should have been about is to describe the domain_resolution_order functionality (you linked the design page in [1]). You also put it among troubleshooting guides, but this is not a troubleshooting guide.

So lets start pretty much from scratch, name the article for example "Domain Resolution Order"  and place it after "Reporting bugs" since it is general information applicable to all providers. 

Multidomain setups as sssd.conf having multiple configured domain is not that common. The domain resolution order was mainly added to allow non-qualified names for trusted AD users (users from subdomains) so this scenario should be used in examples, multidomain setup can be just mentioned.

What the article should describe is:

- what is the default domain order
- by default subdomains users requires fully qualified name
- how to change that behavior by setting domain_resolution_order and how does it affect lookups
- it can also be set globally on ipa server
- etc
-

PR #52, opened 2022-03-01
analyzer: Add clarification to improve documentation

justin-stephenson:


shridhargadekar:
Added information is useful and explains the utility better. Thank you for these additions.

PR #55, opened 2022-03-22
design-pages: added Get and Set per-Component Debug-Level

aplopez:
This Design Page is related to https://github.com/SSSD/sssd/issues/6019

### DIFF
--- a/src/design-pages/sssctl_debug_level.rst
+++ b/src/design-pages/sssctl_debug_level.rst
@@ -1,0 +1,114 @@
+Get and Set per-Component Debug-Level
+#####################################
+
+Related Tickets
+***************
+
+* https://github.com/SSSD/sssd/issues/6019
+
+Problem statement
+*****************
+The debug level can currently be set using the sssctl tool, but it cannot be
+retrieved except by grepping the log file.
+
+In addition, each component can have its own debug level, but the tool sets
+it for all of them and there is no way to limit it to a given component.
+
+Solution overview
+*****************
+Extend the behavior of ``sssctl``’s ``debug-level`` command to:
+
+1. Print the current debug level if no value is provided. Because each
+   component can have a separate debug level, a list will be printed,
+   not just a single, global value.
+
+2. Allow the user to select which component(s) to target. That applies both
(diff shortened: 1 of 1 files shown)


alexey-tikhonov:
@aplopez, very detailed write-up :+1: 

@pbrezina, didn't you recently mention that `sssctl` ought to be re-implemented in python (and hence any large modification of C-based version are undesirable)?

This especially applies to messing with config files. Code to make those manipulations atomic in C won't be neat...

pbrezina:
> @aplopez, very detailed write-up +1
> 
> @pbrezina, didn't you recently mention that `sssctl` ought to be re-implemented in python (and hence any large modification of C-based version is undesirable)?

Yes, `get` operation is not difficult and it is a nice first task for Alejandro.

> This especially applies to messing with config files. Code to make those manipulations atomic in C won't be neat...

Setting the debug level permanently is something we added on top of the original request. I'm not yet familiar with libini write api, If it will be more difficult, we can drop it.

aplopez:
> Setting the debug level permanently is something we added on top of the original request. I'm not yet familiar with libini write api, If it will be more difficult, we can drop it.

Reading and setting the debug level is done using `confdb`, but to make it persistent `libini` must be used in parallel. This seems to me a bit awkward, specially because (if I understood correctly) we need to initialize the library and structures, read the configuration, modify it and write the whole file again, hoping nobody else is doing the same thing at the same time (race condition - I don't know if `libini` locks the file but it would be most unusual).

If the tool will be re-implemented in the near future, I think we can skip the feature.

pbrezina:
Ok, I agree.

aplopez:
Shall I remove the paragraph talking about the --persistent flag?

pbrezina:
Yes, please. Also update it to the latest version (switch to dbus instead of reading confdb).

pbrezina:
Thank you. It looks good, but lets push this after the PR is accepted so further changes can be incorporated if needed.

aplopez:
Updated the design page to reflect the latest implementation.

pbrezina (approved):

PR #57, opened 2022-06-17
design pages: Enable passkey authentication

ikerexxe:
passkey - Authentication in a centralized environment

Relates to https://github.com/SSSD/sssd/issues/6228

### DIFF
--- a/src/design-pages/passkey_authentication.rst
+++ b/src/design-pages/passkey_authentication.rst
@@ -1,0 +1,296 @@
+Passkey - Authentication in a centralized environment
+=======================================================
+
+Related ticket(s):
+------------------
+ * https://github.com/SSSD/sssd/issues/6228
+
+Problem statement
+-----------------
+The use of new tools to authenticate users, such as 2FA, U2F and FIDO2, is
+becoming increasingly popular and currently SSSD doesn’t provide any way to use
+the latter. This diminishes the value provided by SSSD in some environments
+like big companies and governments, where the usage of these authentication
+mechanisms is becoming a common pattern.
+
+SSSD should provide a way to authenticate users that use a passkey. This would
+add more value by attracting the types of users mentioned in the previous
+paragraph.
+
+For the purpose of this feature, passkey is a FIDO2 compatible device supported
+by the libfido2 library. If a hardware token implements other authentication
+mechanisms aside from FIDO2, these aren't considered by this feature.
+
+Use cases
+---------
+As a user, I want to sit in front of a system enrolled into a centralized
+identity management solution and log in (desktop and/or terminal) using a
+passkey device connected to the system.
+
+Solution overview
+------------------------
+The objective is to use a passkey to locally authenticate a user against 
+centralized identity management system. For that purpose an integration with an
+LDAP server like AD or FreeIPA is needed.
+
+First, the key is registered in the LDAP server. Then, the SSSD clients should
+get the user credentials from this server, including the user key mapping data.
+On top of that, SSSD should communicate with the key to get the information
+stored in the device. Once this is done, SSSD should use all this data to
+authenticate the user.
+
+The biggest advantage of this approach is the usage of a central repository to
+store the user's credentials. In the long term this gives more flexibility as
+the users and keys can be authorized and revoked by the system administrator
+easily.
+
+Implementation details
+----------------------
+
+The implementation will loosely follow existing smartcard authentication in
+SSSD. On top of that, and as mentioned before, the information will be stored
+in two places: in the LDAP server and on the passkey. The server will contain
+the public user credentials, including the user key mapping data, while the
+passkey device will store the private key necessary to authenticate the user.
+
+Key registration
+****************
+The first step is to register the user with a key. SSSD provides a helper
+utility to handle the passkeys, and it can be used to create the key mapping
+string (similar to pamu2fcfg). Instead of calling the helper utility from the
+command line the system administrator uses 
+`sssctl <https://docs.pagure.org/sssd.sssd/design_pages/sssctl.html>`__. This
+enables to unify the user experience. If sssctl is run by an unprivileged user,
+then additional udev rules might need to be configured. Please, check the
+instructions from the key provider to know how to add them.
+
+The output of this command can be set to the LDAP server in a dedicated
+attribute. For AD we use the
+`altSecurityIdentities <https://learn.microsoft.com/en-us/windows/win32/adschema/a-altsecurityidentities>`__
+with a prefix (i.e. passkey). For FreeIPA we use a new attribute called
+ipapasskey from the ipapasskeyuser objectclass. For any other LDAP server we
+use the passkey attribute.
+
+The format for server-side credentials for the key mapping is
+``passkey:credentialId,pemPublicKey``.
+
+The format for
+`discoverable credentials <https://developers.yubico.com/WebAuthn/WebAuthn_Developer_Guide/Resident_Keys.html>`__ for the key mapping is
+``passkey:credentialId,pemPublicKey,userId``.
+
+IPA process
++++++++++++
+FreeIPA provides a direct way to register the passkey attributes to the user
+entry. As specified in FreeIPA's documentation this can be done with the
+`ipa user-add-passkey` command or directly in the WebUI. This way FreeIPA can
+take advantage of the SSSD and libfido2 integration to communicate with the
+device in a single step.
+
+AD (and any other LDAP server) process
+++++++++++++++++++++++++++++++++++++++
+The users, or the system administrator, calls sssctl. The user can manually
+add this information to their AD object using AD's 'User and Computers'
+utility, or any other attribute in case of another LDAP server.
+
+LDAP attributes
+***************
+FreeIPA ID provider uses LDAP attribute ``ipaPassKey`` to retrieve the passkey
+mapping data string as part of the user search in the backend. Store this in
+the sysdb. The passkey mapping is obtained during the registration phase.
+Example:
+
+* ipapasskey: passkey:9S87qLk8/RxYJ3skwwYduomAM+/HDtz41N0+w/vRL6aGKJkLMsg+2OhO0E8pK5DuO1KmdK61K8PmH7jiYuOqbg==,9YE1s/f7J47h2A/DXCVFWulqoBXFzCSxcbGEBadkpSUFjwUudhPLnPUTv2qNamakXJgRYCZQ7vpS/t5zXMLnkw==
+
+In the Active Directory case, the
+`altSecurityIdentities <https://learn.microsoft.com/en-us/windows/win32/adschema/a-altsecurityidentities>`__
+LDAP attribute is used to store the passkey mapping data. Example:
+
+* altSecurityIdentities: passkey:9S87qLk8/RxYJ3skwwYduomAM+/HDtz41N0+w/vRL6aGKJkLMsg+2OhO0E8pK5DuO1KmdK61K8PmH7jiYuOqbg==,9YE1s/f7J47h2A/DXCVFWulqoBXFzCSxcbGEBadkpSUFjwUudhPLnPUTv2qNamakXJgRYCZQ7vpS/t5zXMLnkw==
+
+Finally, IPA ``iparequireuserverification`` LDAP attribute is retrieved from
+the IPA passkey configuration. Store this in the sysdb associated with  the IPA
+domain. This attribute stores a string (on, off, default). Example:
+
+* iparequireuserverification: on
+
+User authentication with the key
+********************************
+The idea is to create a helper process similar to the p11_child. The helper
+process receives the passkey’s user key mapping data as an argument alongside
+other parameters (check
+:ref:`Configuration options<configuration-options>` for more information).
+
+This helper process heavily relies on the libfido2 library. The first steps
+are to initialize the library and get the list of keys connected to the system.
+Next the helper process loads the configuration parameters in the assert
+structure, that comprises a collection of statements, each statement has a
+mapping for a challenge, a credential, a signature, and ancillary attributes.
+Finally, the helper process opens the key device and ask it for an
+authentication according to the information provided in the
+`assert <https://developers.yubico.com/libfido2/Manuals/fido_assert_new.html>`__
+structure.
+
+Call the passkey child in the PAM responder
++++++++++++++++++++++++++++++++++++++++++++
+In ``pam_forwarder()`` check if passkey authentication is enabled (if
+pam_passkey_auth boolean option is true and pam cmd == SSS_PAM_AUTHENTICATE).
+If so, lookup the sysdb user object of the user logging in and read the sysdb
+passkey mapping value. Parse out the key-handle and public-key strings from the
+passkey mapping value. Read passkey related configuration options
+(passkey_verification, passkey_child_timeout, debug_libfido2) and set up the
+arguments (Domain, Key-handle and Public-key) for the passkey-child.
+
+Execute the passkey_child passing the arguments listed above. The passkey PIN
+is retrieved from the PAM conversation and written to the stdin of the
+forked passkey child process(Similar to ``get_p11_child_write_buffer()``).
+Check the passkey_child return code, return PAM_SUCCESS or failure based on the
+result and call ``pam_reply()``. If the credential is discoverable, then the
+passkey_child also prints the ``userId``, and the PAM responder has to
+compare it with the one provided by the LDAP server. If they match, then it can
+return PAM_SUCCESS.
+
+Prompting implementation
+************************
+The prompting is handled by the PAM responder. For that purpose, the passkey
+related authtokens is added in ``src/util/authtok.c``,
+``sss_authtok_set_passkey_pin()`` and ``sss_authtok_get_passkey_pin()``.
+
+In ``src/sss_client/pam_sss.c`` add a ``prompt_passkey()`` option which takes
+the prompt as an argument to call ``do_pam_conversation()``. This sets the
+type SSS_AUTHTOK_TYPE_PASSKEY_PIN.
+
+In the PAM responder, ``pam_set_passkey_prompting_options()`` is added to
+``src/responder/pam/pam_prompting_config.c`` to check and handle the PAM
+prompting passkey configuration options.
+
+``pc_list_add_passkey()`` copies the interactive prompt message to
+``pc->data.passkey.prompt_inter`` and the touch prompt message to
+``pc->data.passkey.prompt_touch`` and set type PC_TYPE_PASSKEY.
+
+If interactive or touch options are set to false, then these values are set to
+an empty string. If pam_sss reads an empty string for these prompts it does not
+include them in the PAM conversation. This is done because we can't
+fallback to a default prompt message in this situation, we need to skip the
+interactive or touch prompt entirely.
+
+Add the prompt message to the data buffer response back to pam_sss by adding
+PC_TYPE_PASSKEY to case statements in ``pam_get_response_prompt_config()`` and
+``pc_list_from_response()``. 
+
+In ``prompt_passkey()``, 1 to 3 messages are provided to the PAM conversation
+function. The first “Enter PIN:” PAM message is always created. If interactive
+and/or touch prompts are enabled in the prompt configuration then those
+messages are added to the pam message array and provided to the pam
+conversation.
+
+.. _configuration-options:
+
+Configuration options
+*********************
+"pam_passkey_auth" enables the passkey device authentication.
+
+"passkey_verification" is added to the SSSD configuration options. It is
+similar to the “certificate_verification” option for the p11_child, as it
+contains the parameters needed to tune the passkey_child helper
+process. The list of parameters and their meaning:
+
+* user_verification: if set to true, requires user verification (i.e. PIN,
+  password) during authentication. If set to false, does not request user
+  verification during authentication. The default is that the key itself
+  decides what to do.
+
+"passkey_child_timeout" sets the timeout for the PAM responder to wait for
+passkey_child to finish.
+
+"passkey_debug_libfido2" prints libfido2 library messages. It's under the
+``[pam]`` section and it defaults to false.
+
+Prompting Configuration
++++++++++++++++++++++++
+Another section called [prompting/passkey] is added. This section is similar
+to other prompting sections (i.e. 2fa). The list of options and their meaning:
+
+* interactive: set to prompt a message and wait before testing the presence of
+  a passkey device. Recommended if your device doesn’t have a tactile trigger.
+
+* interactive_prompt=your prompt here: set individual prompt message for
+  interactive mode. Default is: “Insert your Passkey device, then press ENTER.”
+
+* touch: set to prompt a message to remind the user to touch the device.
+
+* touch_prompt=your prompt here: set individual prompt message for the cue
+  option. Default is: “Please touch the device."
+
+An example prompt interaction can look like: ::
+
+    Insert your Passkey device, then press ENTER.
+    < ENTER >
+    Enter PIN: 1234
+    Please touch the device.
+
+User verification
++++++++++++++++++
+The user verification can be set in various places:
+
+* In the IPA passkey configuration
+* In the local sssd.conf
+
+IPA passkey configuration user verification requirement overrides local
+sssd.conf.
+
+Registration process
+--------------------
+
+sssctl
+******
+
+Each passkey needs to be registered before it can be used for authentication.
+This registration process is quite simple; the user connects the hardware
+token to the computer, and then, executes the
+``sssctl passkey-exec  --register`` command.
+
+The command contains several parameters that slightly change its behaviour or
+its output. The following is a list of the mandatory options:
+
+* domain: LDAP domain name. Also know as
+  `Relying Party <https://www.w3.org/TR/webauthn-2/#webauthn-relying-party>`__
+  in the WebAuthn standard.
+
+List of optional parameters:
+
+* username: this is the username as registered in the LDAP server.
+
+* type: public key cryptography. There are three possible types: es256, rs256
+  and EdDSA. The default is es256.
+
+* user-verification: requires user verification (i.e. PIN, fingerprint).
+
+* cred-type: credential type (server-side or discoverable).
+
+* debug-libfido2: flag to print libfido2 library messages.
+
+The most basic example of a registration would be the following: ::
+
+    # sssctl passkey-exec --register --username=USERNAME --domain=DOMAIN
+
+This outputs the key mapping data (
+``passkey:credentialId,pemPublicKey,userId``) that is used as the input for the
+registration in the LDAP server. In AD and other LDAP servers the output is
+copied to the LDAP attribute. In FreeIPA, the key mapping can copied to the
+WebUI or to a command:
+``ipa user-add-passkey USERNAME KEY_MAPPING``, or you can use the FreeIPA's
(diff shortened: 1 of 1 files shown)


abbra:
In the design page there is no explanation how registration process happens and what parameters used. It took me some time to find out that a relying party, for example, is the name of the SSSD domain user belongs to. I think these details should be part of the documentation.

ikerexxe:
> In the design page there is no explanation how registration process happens and what parameters used. It took me some time to find out that a relying party, for example, is the name of the SSSD domain user belongs to. I think these details should be part of the documentation.

I've added a new section, `Registration process`, to explain it.

src/design-pages/passkey_authentication.rst
```diff
@@ -0,0 +1,313 @@
...
+---------
+As a user, I want to walk to any system on a network and log in (desktop and/or
+terminal) using a passkey device connected to the system. The user credentials
+will be stored in a central management system like FreeIPA or Active Directory.
+
+Solution overview
+------------------------
+The objective is to use a passkey to locally authenticate a user against 
+centralized identity management system. For that purpose an integration with an
+LDAP server like AD or FreeIPA is needed. SSSD should get the user credentials
```
justin-stephenson, line 33: s/SSSD/SSSD clients/

Also, I would explicitly state that first the passkey data needs to be added to the LDAP directory server.

src/design-pages/passkey_authentication.rst
```diff
@@ -0,0 +1,313 @@
...
+factor can also be used for doing a direct LDAP bind. This will not be possible
+with the new passkey scheme. Direct LDAP bind support will not be in the scope
+of this implementation.
+
+Implementation details
+----------------------
+
+The implementation will loosely follow existing smartcard authentication in
+SSSD. On top of that, and as mentioned before, the information will be stored
+in two places: in the LDAP server and in the passkey. The server will contain
```
justin-stephenson, line 70: s/in the passkey/on the passkey/  sounds a bit better IMO

src/design-pages/passkey_authentication.rst
```diff
@@ -0,0 +1,313 @@
...
+``src/providers/ipa/ipa_opts.c`` to retrieve the passkey mapping data
+string as part of the user search in the backend. Store this in the sysdb. The
+passkey mapping is obtained during the registration phase. Example:
+
+* ipapasskey: passkey:9S87qLk8/RxYJ3skwwYduomAM+/HDtz41N0+w/vRL6aGKJkLMsg+2OhO0E8pK5DuO1KmdK61K8PmH7jiYuOqbg==,9YE1s/f7J47h2A/DXCVFWulqoBXFzCSxcbGEBadkpSUFjwUudhPLnPUTv2qNamakXJgRYCZQ7vpS/t5zXMLnkw==
+
+In the Active Directory case, the
+`altSecurityIdentities <https://learn.microsoft.com/en-us/windows/win32/adschema/a-altsecurityidentities>`__
+LDAP attribute will be used to store the passkey mapping data. Add this to
+``src/proviers/ad/ad_opts.c`` to retrieve from the AD user object. Example:
```
justin-stephenson, line 126: Probably this line is not needed about `src/providers/ad`ad_opts.c`

src/design-pages/passkey_authentication.rst
```diff
@@ -0,0 +1,313 @@
...
+be similar to the “certificate_verification” option for the p11_child, as it
+will contain the parameters needed to tune the passkey_child helper
+process. The list of parameters and their meaning:
+
+* user-verification: if set to true, requires user verification (i.e. PIN,
+  password) during authentication. If set to false, does not request user
+  verification during authentication. The default is that the key itself
+  decides what to do.
+
+* debug-libfido2: flag to print libfido2 library messages.
```
justin-stephenson, line 220: defaults to false

src/design-pages/passkey_authentication.rst
```diff
@@ -0,0 +1,313 @@
...
+
+* touch: set to prompt a message to remind the user to touch the device.
+
+* touch-prompt=your prompt here: set individual prompt message for the cue
+  option. Default will be: “Please touch the device."
+
+An example prompt interaction can look like: ::
+
+    Insert your Passkey device, then press ENTER. 
+    ENTER
```
justin-stephenson, line 245: <ENTER> to signify it is not text but a keyboard 'enter'  being pressed

src/design-pages/passkey_authentication.rst
ikerexxe: I did both things

src/design-pages/passkey_authentication.rst
ikerexxe: Yeah, this is a recurrent problem for Spanish-speaking, we fail when using in/on/at. Thanks for catching it.

ikerexxe:
@justin-stephenson I fixed the comments and I added `If the credential is discoverable, then the passkey_child will also print the userId, and the PAM responder has to compare it with the one provided by the LDAP server. If they match then it can return PAM_SUCCESS.` in the `Call the passkey child in the PAM responder` section.

src/design-pages/passkey_authentication.rst
justin-stephenson: Thanks, LGTM.

andreboscatto:
@abbra it would be nice if you could take a look :)

abbra:
@andreboscatto I will, I need to get through this flu first...

pbrezina:
Hi Iker, can you please squash the commits together? It will make it easier to provide comments.

src/design-pages/passkey_authentication.rst
```diff
@@ -0,0 +1,318 @@
...
+
+Key registration
+****************
+The first step is to register the user with a key. SSSD will provide a helper
+utility to handle the passkeys, and it can be used to create the key mapping
+string (similar to pamu2fcfg). Instead of calling the helper utility from the
+command line the system administrator would use 
+`sssctl <https://docs.pagure.org/sssd.sssd/design_pages/sssctl.html>`__. This
+would enable to unify the user experience. Then, the output of this command can
+be set to the LDAP server in a dedicated attribute. For AD we could use the
```
abbra, line 85: is the attribute choice decided already for each ID provider? If so, please replace 'we could use' by a more definitive statement that lists the attributes per each ID provider.

src/design-pages/passkey_authentication.rst
```diff
@@ -0,0 +1,318 @@
...
+for the key mapping will be
+``passkey:credentialId,pemPublicKey``.
+
+The format for
+`discoverable credentials <https://developers.yubico.com/WebAuthn/WebAuthn_Developer_Guide/Resident_Keys.html>`__ for the key mapping will be
+``passkey:credentialId,pemPublicKey,userId``.
+
+IPA process
++++++++++++
+FreeIPA would provide a direct way to register the passkey attributes to the
```
abbra, line 102: I think for IPA you should provide a link to readthedocs page. Since IPA part is not merged yet (chicken and egg problem with SSSD release with passkey yet), we should probably update this part of the design once it is published. It is definitely good to replace 'would provide' and below 'can be done' by more concrete statement based on the IPA design.  Please see IPA design here: https://github.com/flo-renaud/freeipa/blob/passkey/doc/designs/passkeys.md

src/design-pages/passkey_authentication.rst
```diff
@@ -0,0 +1,318 @@
...
+
+AD (and any other LDAP server) process
+++++++++++++++++++++++++++++++++++++++
+The users, or the system administrator, would call sssctl. The user can
+manually add this information to their AD object using AD's 'User and
+Computers' utility, or any other attribute in case of another LDAP server.
+
+LDAP attributes
+***************
+Newly added IPA ``ipapasskey`` LDAP attribute will be added to
```
abbra, line 119: Please rephrase as factual statement. E.g. 'FreeIPA ID provider will use LDAP attribute ``ipaPassKey`` to retrieve the passkey mapping data ...".

src/design-pages/passkey_authentication.rst
```diff
@@ -0,0 +1,318 @@
...
+
+* ipapasskey: passkey:9S87qLk8/RxYJ3skwwYduomAM+/HDtz41N0+w/vRL6aGKJkLMsg+2OhO0E8pK5DuO1KmdK61K8PmH7jiYuOqbg==,9YE1s/f7J47h2A/DXCVFWulqoBXFzCSxcbGEBadkpSUFjwUudhPLnPUTv2qNamakXJgRYCZQ7vpS/t5zXMLnkw==
+
+In the Active Directory case, the
+`altSecurityIdentities <https://learn.microsoft.com/en-us/windows/win32/adschema/a-altsecurityidentities>`__
+LDAP attribute will be used to store the passkey mapping data. Example:
+
+* altSecurityIdentities: passkey:9S87qLk8/RxYJ3skwwYduomAM+/HDtz41N0+w/vRL6aGKJkLMsg+2OhO0E8pK5DuO1KmdK61K8PmH7jiYuOqbg==,9YE1s/f7J47h2A/DXCVFWulqoBXFzCSxcbGEBadkpSUFjwUudhPLnPUTv2qNamakXJgRYCZQ7vpS/t5zXMLnkw==
+
+Finally, newly added IPA ``iparequireuserverification`` LDAP attribute will be
```
abbra, line 132: Same here. Please rephrase -- no need for 'newly added ... LDAP attribute', it will not be new by the time it is merged in IPA already, just state the attribute name.

src/design-pages/passkey_authentication.rst
```diff
@@ -0,0 +1,318 @@
+Passkey - Authentication in a centralized environment
```
abbra, line 1: A general observation. Now that we have most components implemented, it would make sense to rephrase those parts of the design that speak in potential way to state facts. E.g. instead of "will use X" or "would use X", just state the choice, "implementation uses X".

src/design-pages/passkey_authentication.rst
```diff
@@ -0,0 +1,318 @@
...
+Finally, newly added IPA ``iparequireuserverification`` LDAP attribute will be
+retrieved from the IPA passkey configuration. Store this in the sysdb
+associated with the IPA domain. This attribute stores a string (on, off,
+default). Example:
+
+* iparequireuserverification: on
+
+User authentication with the key
+********************************
+The idea is to create a helper process similar to the p11_child. The helper
```
abbra, line 141: Please rephrase -- we already have implementation, this is not an idea anymore.

src/design-pages/passkey_authentication.rst
```diff
@@ -0,0 +1,318 @@
...
++++++++++++++++++++++++++++++++++++++++++++
+In ``pam_forwarder()`` check if passkey authentication is enabled (if
+pam_passkey_auth boolean option is true and pam cmd == SSS_PAM_AUTHENTICATE).
+If so, lookup the sysdb user object of the user logging in and read the sysdb
+passkey mapping value. Parse out the key-handle and public-key strings from the
+passkey mapping value. Read passkey related configuration options
+(passkey_verification, passkey_child_timeout, debug_libfido2) and set up the
+arguments (Domain, Key-handle and Public-key) for the passkey-child.
+
+Execute the passkey_child passing the arguments listed above. The passkey PIN
```
abbra, line 166: It is unclear under what user the passkey_child is launched. Since passkeys often require access to USB or other external bus to communicate, it would be good to mention how to achieve that (udev rules). Specifically, this should be covered if passkey_child does not run as root.

abbra (changes requested):
Hi @ikerexxe, thank you for the design document. In general, it is good one already. I just left few comments that ask for changes since we already have made certain changes since this document was written and it would be better to simply state the decisions rather than list the choices that weren't yet finalized when the document was written.

src/design-pages/passkey_authentication.rst
ikerexxe: I'm keeping this open until until IPA part is merged.

src/design-pages/passkey_authentication.rst
ikerexxe: I explain this at the top of the document, where I explain how to use the `sssctl` command for the first time.

ikerexxe:
@abbra thanks for the feedback. As a general note, I have rephrased all potential statements (will/would) to affirmative phrases as requested.

src/design-pages/passkey_authentication.rst
```diff
@@ -0,0 +1,318 @@
...
+paragraph.
+
+For the purpose of this feature, passkey is a FIDO2 compatible device supported
+by the libfido2 library.
+
+Use cases
+---------
+As a user, I want to walk to any system on a network and log in (desktop and/or
+terminal) using a passkey device connected to the system. The user credentials
+will be stored in a central management system like FreeIPA or Active Directory.
```
pbrezina, line 27: ATM We only support local login over SSSD+FIDO. We don't support SSH+SSSD+FIDO, right? Therefore "I want to walk to any system on a network" is not part of this design page.

src/design-pages/passkey_authentication.rst
```diff
@@ -0,0 +1,318 @@
...
+The approach explained in this document relies on the 
+`FIDO2 <https://en.wikipedia.org/wiki/FIDO2_Project>`__ standard, which uses a
+public-private key scheme to perform the authentication checks on the client.
+Besides, the solution is general enough to accept different LDAP servers like
+AD and IPA.
+
+Another difference is that with the current IPA feature, the first and second
+factor can also be used for doing a direct LDAP bind. This will not be possible
+with the new passkey scheme. Direct LDAP bind support will not be in the scope
+of this implementation.
```
pbrezina, line 66: I found this paragraph confusing. Is direct bind with FIDO supported on IPA but unsupported on plain LDAP? Or do you refer to OTP?

src/design-pages/passkey_authentication.rst
```diff
@@ -0,0 +1,318 @@
...
+
+The output of this command can be set to the LDAP server in a dedicated
+attribute. For AD we use the
+`altSecurityIdentities <https://learn.microsoft.com/en-us/windows/win32/adschema/a-altsecurityidentities>`__
+with a prefix (i.e. passkey). For FreeIPA we use a new attribute called
+ipapasskey from the ipapasskeyuser objectclass. For any other LDAP server we
+use the passkey attribute.
+
+The format for
+`server-side credentials <https://developers.yubico.com/WebAuthn/WebAuthn_Developer_Guide/Resident_Keys.html>`__
```
pbrezina, line 96: Are you sure this link is valid? It is the same as the one for discoverable credentials and does not explain server-side creds.

src/design-pages/passkey_authentication.rst
```diff
@@ -0,0 +1,318 @@
...
+enables to unify the user experience. If sssctl is run by an unprivileged user,
+then additional udev rules might need to be configured. Please, check the
+instructions from the key provider to know how to add them.
+
+The output of this command can be set to the LDAP server in a dedicated
+attribute. For AD we use the
+`altSecurityIdentities <https://learn.microsoft.com/en-us/windows/win32/adschema/a-altsecurityidentities>`__
+with a prefix (i.e. passkey). For FreeIPA we use a new attribute called
+ipapasskey from the ipapasskeyuser objectclass. For any other LDAP server we
+use the passkey attribute.
```
pbrezina, line 93: Provide a link to the passkey ldap schema?

src/design-pages/passkey_authentication.rst
```diff
@@ -0,0 +1,318 @@
...
+`discoverable credentials <https://developers.yubico.com/WebAuthn/WebAuthn_Developer_Guide/Resident_Keys.html>`__ for the key mapping is
+``passkey:credentialId,pemPublicKey,userId``.
+
+IPA process
++++++++++++
+FreeIPA provides a direct way to register the passkey attributes to the user
+entry. As specified in FreeIPA's documentation this can be done with the
+`user-add-passkey` command or directly in the WebUI. This way FreeIPA can take
+advantage of the SSSD and libfido2 integration to communicate with the
+device in a single step.
```
pbrezina, line 110: So you either call sssctl + user-add-passkey, or you use webui which talks to the fido2 device directly? Or how does it work?

src/design-pages/passkey_authentication.rst
```diff
@@ -0,0 +1,318 @@
...
+
+AD (and any other LDAP server) process
+++++++++++++++++++++++++++++++++++++++
+The users, or the system administrator, calls sssctl. The user can manually
+add this information to their AD object using AD's 'User and Computers'
+utility, or any other attribute in case of another LDAP server.
```
pbrezina, line 116: The administrator must have temporary access to the fido2 device, right?

src/design-pages/passkey_authentication.rst
```diff
@@ -0,0 +1,318 @@
...
+arguments (Domain, Key-handle and Public-key) for the passkey-child.
+
+Execute the passkey_child passing the arguments listed above. The passkey PIN
+is retrieved from the PAM conversation and written to the stdin of the
+forked passkey child process(Similar to ``get_p11_child_write_buffer()``).
+Check the passkey_child return code, return PAM_SUCCESS or failure based on the
+result and call ``pam_reply()``. If the credential is discoverable, then the
+passkey_child also prints the ``userId``, and the PAM responder has to
+compare it with the one provided by the LDAP server. If they match, then it can
+return PAM_SUCCESS.
```
pbrezina, line 173: Comparing of userId might be better to do in passkey_child so the whole logic in in single process.

src/design-pages/passkey_authentication.rst
```diff
@@ -0,0 +1,318 @@
...
+
+* user-verification: requires user verification (i.e. PIN, fingerprint).
+
+* cred-type: credential type (server-side or discoverable).
+
+* debug-libfido2: flag to print libfido2 library messages.
+
+The most basic example of a registration would be the following: ::
+
+    # sssctl passkey-exec --register --username=USERNAME --domain=DOMAIN
```
pbrezina, line 296: I know I'm late and I haven't reviewed patches that already land in, but I'm not very fond of current implementation.

```
errno_t sssctl_passkey_exec(struct sss_cmdline *cmdline,
                            struct sss_tool_ctx *tool_ctx,
                            void *pvt)
{
    errno_t ret;

    ret = sssctl_wrap_command(SSS_PASSKEY_CHILD, cmdline, tool_ctx, pvt);

    return ret;
}
```

Wouldn't it be better to provide `sssctl passkey-register` command instead of allowing to call anything in passkey_child?

pbrezina (commented):
Hi Iker, see few comments inside.

src/design-pages/passkey_authentication.rst
ikerexxe: Right, we only support SSSD+FIDO. The part of the phrase that you are quoting is exactly the one that specifies that you need to be physically near the system to perform the authentication, you can't do it remotely by SSH or any other means.

pbrezina: Please, rephrase it to make it more clear. To me, "walk to any system on a network" means to ssh to any system on a network.

abbra: I think you should drop 'any system on a network'  and replace it with something like 'a system enrolled into a centralized identity management solution', this would be more clear.

src/design-pages/passkey_authentication.rst
abbra: I think this paragraph can be dropped, it does not add anything to clarify the implementation.

ikerexxe: Direct bind isn't supported in the scope of this feature. In this section of the document I'm explaining the differences between the already implement OTP feature and this one.

It's kind of confusing because the FIDO2 keys provided by several vendors also contain other protocols (OTP, PIV, etc.). So, even if we refer to them as FIDO2 keys, they provide FIDO2 and other mechanisms. When I started writing this document and asked for feedback, some people didn't understand why the actual implementation (OTP) wasn't good enough. Our users were already using FIDO2 keys to authenticate, why did we need another implementation? That's why I wrote this section.

abbra: I think what is needed here is an explicit statement that only FIDO2 part is supported here. If a hardware token implements other authentication mechanisms aside from FIDO2, these aren't considered by the passkey authentication feature. What is done by the FreeIPA server side should not be documented here.

src/design-pages/passkey_authentication.rst
ikerexxe: Yes, that sounds good to me. I'll keep this open until the FreeIPA changes are merged in the "central" repository and then I'll add the link.

src/design-pages/passkey_authentication.rst
ikerexxe: Yes

src/design-pages/passkey_authentication.rst
ikerexxe: @sumit-bose I remember you asked to do the comparison in the PAM responder, but I can't remember why. Can you enlighten us?

src/design-pages/passkey_authentication.rst
ikerexxe: That makes sense. @justin-stephenson what do you think?

src/design-pages/passkey_authentication.rst
ikerexxe: I rephrased it to take into account your feedback

src/design-pages/passkey_authentication.rst
ikerexxe: I dropped this section from the document as it doesn't make sense in this context and I added an explanation in the `Problem statement` to indicate that his feature is FIDO2 only.

src/design-pages/passkey_authentication.rst
ikerexxe: It's `ipa user-add-passkey` or WebUI. I've added the `ipa` command to make it more clear.

src/design-pages/passkey_authentication.rst
sumit-bose: Hi,

the idea was, since the userID is created as a random number, ot handle it as a second shared secret so that e.g. in the FreeIPA Kerberos case the userID is only know to the server and the token. The userID will not be send to the client by the server but only returned fom the client to the server so that the server can check if userID is the same as the stored one. Similar for the local authentication, passkey_child should not know which value is expected here.

Originally I thought that the userId will be part of the authenticator data returned by the token and hence is protected against modification by the signature. But it looks it can only be send separately. Not sure if it would need some additional protection in this case?

bye,
Sumit

src/design-pages/passkey_authentication.rst
pbrezina: But users need to call only register, right? You can call passkey_child directly for debugging and testing purpose.

ikerexxe: Opened https://issues.redhat.com/browse/SSSD-5644 to track this work.

src/design-pages/passkey_authentication.rst
ikerexxe: As discussed in the meeting this part of the design is fine. So I'm resolving this conversation.

src/design-pages/passkey_authentication.rst
```diff
@@ -0,0 +1,295 @@
...
+similar to the “certificate_verification” option for the p11_child, as it
+contains the parameters needed to tune the passkey_child helper
+process. The list of parameters and their meaning:
+
+* user-verification: if set to true, requires user verification (i.e. PIN,
+  password) during authentication. If set to false, does not request user
+  verification during authentication. The default is that the key itself
+  decides what to do.
+
+* debug-libfido2: flag to print libfido2 library messages. Defaults to false.
```
pbrezina, line 202: Should this really be inside passkey_verification option?  Wouldn't a standalone option be better?

src/design-pages/passkey_authentication.rst
```diff
@@ -0,0 +1,295 @@
...
+
+Prompting Configuration
++++++++++++++++++++++++
+Another section called [prompting/passkey] is added. This section is similar
+to other prompting sections (i.e. 2fa). The list of options and their meaning:
+
+* interactive: set to prompt a message and wait before testing the presence of
+  a passkey device. Recommended if your device doesn’t have a tactile trigger.
+
+* inter-prompt=your prompt here: set individual prompt message for interactive
```
pbrezina, line 215: Underscore instead of dash?

src/design-pages/passkey_authentication.rst
```diff
@@ -0,0 +1,295 @@
...
+  mode. Default is: “Insert your Passkey device, then press ENTER.”
+
+* touch: set to prompt a message to remind the user to touch the device.
+
+* touch-prompt=your prompt here: set individual prompt message for the cue
```
pbrezina, line 220: Underscore instead of dash?

src/design-pages/passkey_authentication.rst
ikerexxe: I'd like to keep the configuration behaviour as similar to the `p11_child` as possible. Changing it to a stand-alone option would do the contrary.

src/design-pages/passkey_authentication.rst
ikerexxe: Right, and it's `interactive_prompt` instead of `inter_prompt` so I've also updated that

src/design-pages/passkey_authentication.rst
ikerexxe: Fixed

src/design-pages/passkey_authentication.rst
pbrezina: I'm sorry, but this is debugging, it does not have anything to do with verification so it does not belong there. What similarities are you referring to? I don't see any debug value in `certificate_verification`.

ikerexxe: I meant, that all the configuration parameters for `p11_child` are defined in `certificate_verification`, and I'd like to do the same for passkey. It's true though that this is a debug parameter. Are you suggesting to create a new passkey section and move it to its own configuration option?

pbrezina: You can create passkey_debug_libfido2 inside [pam].

ikerexxe: Done. Justin will take of changing the implementation to match this change.

src/design-pages/passkey_authentication.rst
```diff
@@ -0,0 +1,295 @@
...
+
+* user_verification: if set to true, requires user verification (i.e. PIN,
+  password) during authentication. If set to false, does not request user
+  verification during authentication. The default is that the key itself
+  decides what to do.
+
+"passkey_child_timeout" sets the timeout for the PAM responder to wait for
+passkey_child to finish.
+
+"debug_libfido2" prints libfido2 library messages. Defaults to false.
```
justin-stephenson, line 205: This will need to be updated to indicate it is now under the [pam] section, and I renamed it to `passkey_debug_libfido2`

ikerexxe: Done, and thanks for taking care of it!

pbrezina (approved):
Thank you, ack.

abbra:
LGTM. The Github UI is not allowing to add approval.

URLs should be updated once IPA merged the PR for passkeys which will only happen after SSSD merged both phases.

src/design-pages/passkey_authentication.rst
ikerexxe: Opened https://github.com/SSSD/sssd.io/pull/69 to include the link

src/design-pages/passkey_authentication.rst
ikerexxe: Opened https://github.com/SSSD/sssd.io/pull/69 to include the link

PR #59, opened 2022-07-13
design-pages: redesign for id-overrides

sumit-bose:
This is still work-in-progress

### DIFF
--- a/src/design-pages/idoverride_redesign.rst
+++ b/src/design-pages/idoverride_redesign.rst
@@ -1,0 +1,400 @@
+ID-Override - Re-Design
+#######################
+
+Related Tickets
+***************
+
+* `Infopipe ListByCertificate does not return users if more than 1 override matches`_
+* `[RFE] Need an option for sss_override so that uidnumber can be added locally even if it is not available on LDAP server`_
+* `Unexpected interaction between overriding the primary group and the 'auto_private_groups = true' option`_
+* `RFE: Permanently Persistent Local (on-system) SSS Overrides both automating importation (such as after the SSS DB is cleared and SSSD restarted) and a standard location for the override export files`_
+* `IPA in trust. Client not seeing user with override set on AD user`_
+* `User information is being removed from SSSD local Override database`_
+
+.. _Infopipe ListByCertificate does not return users if more than 1 override matches: https://github.com/SSSD/sssd/issues/4456
+.. _[RFE] Need an option for sss_override so that uidnumber can be added locally even if it is not available on LDAP server: https://bugzilla.redhat.com/show_bug.cgi?id=1896728
+.. _Unexpected interaction between overriding the primary group and the 'auto_private_groups = true' option: https://bugzilla.redhat.com/show_bug.cgi?id=1921315
+.. _RFE\: Permanently Persistent Local (on-system) SSS Overrides both automating importation (such as after the SSS DB is cleared and SSSD restarted) and a standard location for the override export files: https://bugzilla.redhat.com/show_bug.cgi?id=1939527
+.. _IPA in trust. Client not seeing user with override set on AD user: https://bugzilla.redhat.com/show_bug.cgi?id=2010512
+.. _User information is being removed from SSSD local Override database: https://bugzilla.redhat.com/show_bug.cgi?id=1452883
+
+
+Problem statement
+*****************
+The original design for id-overrides is described in
+https://www.freeipa.org/page/V4/Migrating_existing_environments_to_Trust and
+all the features described there should of course work as expected after the
+redesign as well. Only the implementation should be updated with performance
+and reliability in mind.
+
+Initial the id-overrides were planned as a feature for the IPA provider. Some
+time later the local id-overrides were added keeping the way how id-overrides
+were handled by the ipa provider. This meant that when the id-override was
+added the related user was looked up as well and added to the cache together
+with the id-override data. And if the user was removed from the cache for
+whatever reasons the matching id-override was removed from the cache as well.
+This works well for well for IPA since there is a central storage for all
+current id-overrides on the IPA servers. But the local id-override data is
+lost and if the user is added to the cache again there will be no id-override
+data to apply. So it would make sense to decouple the id-override data in
+SSSD's cache form the actual user and group objects.
+
+The id-override data is typically stored in a separate tree/container in the
+cache. There is an exception for the "Default Trust View" on IPA servers where
+the id-override data is directly applied to the user or group objects in the
+cache (see original design for the reason) but with the new design it should be
+possibel to get rid of this special handling. So instead of removing the whole
+id-override object automatically only the back-reference to the user or group
+object can be removed and the id-override object is only removed if the source
+object is removed or the local id-override is explicitly removed.
+
+This decoupling would also allow to handle the different change times of
+id-overrides, user and group objects more efficiently. While changes to groups
+are expected to happen quite often by adding or removing member, changes to
+user objects should be less often. Id-overrides are typically created once and
+will most probably never change. Currently, to look a user, e.g. by name,
+there will be first and LDAP search for an id-override with the given name.
+Depending on if an id-override was found or not the user is looked up by the
+id-override anchor or by name respectively. If no id-override was found before
+another search for the id-override is done with the anchor to see if other
+attributes than the name is overwritten. For users or group with no
+id-overrides this means three LDAP searches.
+
+If the id-override data is downloaded once into the cache this can be handled
+more efficiently because the id-override data is already present on the local
+client and can be kept up-to-data in the similar scheme as sudo rules. For
+local id-overrides it would allow to bulk-load many id-overrides into the
+cache in a single run without looking up each corresponding user and group.
+Finally it would make it easier for other providers to implement id-override
+support because relating the id-override and the original object now happens
+inside the SSSD cache and does not involve a complicate sequence in the lookup
+code of the backend.
+
+
+Currently there is a requirement that when looking up attributes in
+id-overrides there should be only one result. While this is true for the most
+attribute it is e.g not true for searches by certificate. There are valid
+use-cases where a single certificate from a Smartcard can be used to log into
+different accounts.
+
+Finally, using id-overrides together with the domain option
+'auto_private_groups' with its different values has to be carefully evaluated
+and it should be clearly defined under which conditions a real group with a
+GID matching the user's GID must exist and when it will be generated
+automatically as user-private-group.
+
+
+
+
+Solution overview
+*****************
+
+Implementation details
+**********************
+
+
+Bulk-loading id-overrides into the cache
+========================================
+While the id-override objects can be updates at runtime individually (although
+we expect that they will rarely change at all) and new id-override object can
+be added one after the other it might be necessary to load a large amount of
+objects at startup if the cache is empty.
+
+In the past we came across various issues with the files provider when loading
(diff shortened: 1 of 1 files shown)


src/design-pages/idoverride_redesign.rst
```diff
@@ -0,0 +1,400 @@
...
+current id-overrides on the IPA servers. But the local id-override data is
+lost and if the user is added to the cache again there will be no id-override
+data to apply. So it would make sense to decouple the id-override data in
+SSSD's cache form the actual user and group objects.
+
+The id-override data is typically stored in a separate tree/container in the
+cache. There is an exception for the "Default Trust View" on IPA servers where
+the id-override data is directly applied to the user or group objects in the
+cache (see original design for the reason) but with the new design it should be
+possibel to get rid of this special handling. So instead of removing the whole
```
jakub-vavra-cz, line 46: typo s/possibel/possible/

src/design-pages/idoverride_redesign.rst
```diff
@@ -0,0 +1,400 @@
...
+* `IPA in trust. Client not seeing user with override set on AD user`_
+* `User information is being removed from SSSD local Override database`_
+
+.. _Infopipe ListByCertificate does not return users if more than 1 override matches: https://github.com/SSSD/sssd/issues/4456
+.. _[RFE] Need an option for sss_override so that uidnumber can be added locally even if it is not available on LDAP server: https://bugzilla.redhat.com/show_bug.cgi?id=1896728
+.. _Unexpected interaction between overriding the primary group and the 'auto_private_groups = true' option: https://bugzilla.redhat.com/show_bug.cgi?id=1921315
+.. _RFE\: Permanently Persistent Local (on-system) SSS Overrides both automating importation (such as after the SSS DB is cleared and SSSD restarted) and a standard location for the override export files: https://bugzilla.redhat.com/show_bug.cgi?id=1939527
+.. _IPA in trust. Client not seeing user with override set on AD user: https://bugzilla.redhat.com/show_bug.cgi?id=2010512
+.. _User information is being removed from SSSD local Override database: https://bugzilla.redhat.com/show_bug.cgi?id=1452883
```
ikerexxe, line 19: Isn't this information duplicated?

src/design-pages/idoverride_redesign.rst
```diff
@@ -0,0 +1,400 @@
...
+
+
+Problem statement
+*****************
+The original design for id-overrides is described in
+https://www.freeipa.org/page/V4/Migrating_existing_environments_to_Trust and
+all the features described there should of course work as expected after the
```
ikerexxe, line 26: s/should of course work/should, of course, work

src/design-pages/idoverride_redesign.rst
```diff
@@ -0,0 +1,400 @@
...
+redesign as well. Only the implementation should be updated with performance
```
ikerexxe, line 27: I would remove the `as well` part

src/design-pages/idoverride_redesign.rst
```diff
@@ -0,0 +1,400 @@
...
+and reliability in mind.
+
+Initial the id-overrides were planned as a feature for the IPA provider. Some
```
ikerexxe, line 30: The initial? Initially? I don't know what's wrong there but I'd definitely change it

src/design-pages/idoverride_redesign.rst
```diff
@@ -0,0 +1,400 @@
...
+time later the local id-overrides were added keeping the way how id-overrides
+were handled by the ipa provider. This meant that when the id-override was
+added the related user was looked up as well and added to the cache together
+with the id-override data. And if the user was removed from the cache for
+whatever reasons the matching id-override was removed from the cache as well.
+This works well for well for IPA since there is a central storage for all
+current id-overrides on the IPA servers. But the local id-override data is
+lost and if the user is added to the cache again there will be no id-override
+data to apply. So it would make sense to decouple the id-override data in
+SSSD's cache form the actual user and group objects.
```
ikerexxe, line 40: s/form/from

src/design-pages/idoverride_redesign.rst
```diff
@@ -0,0 +1,400 @@
...
+
+The id-override data is typically stored in a separate tree/container in the
+cache. There is an exception for the "Default Trust View" on IPA servers where
+the id-override data is directly applied to the user or group objects in the
+cache (see original design for the reason) but with the new design it should be
+possibel to get rid of this special handling. So instead of removing the whole
+id-override object automatically only the back-reference to the user or group
+object can be removed and the id-override object is only removed if the source
+object is removed or the local id-override is explicitly removed.
```
ikerexxe, line 49: I would split this sentence in two or three. Let me try to give you an example but please feel free to do it in another way:

So instead of removing the whole id-override object automatically. only the back-reference to the user or group object is removed. The id-override object is deleted only if the source object is removed or the local id-override is explicitly removed.

src/design-pages/idoverride_redesign.rst
```diff
@@ -0,0 +1,400 @@
...
+there will be first and LDAP search for an id-override with the given name.
+Depending on if an id-override was found or not the user is looked up by the
+id-override anchor or by name respectively. If no id-override was found before
+another search for the id-override is done with the anchor to see if other
+attributes than the name is overwritten. For users or group with no
+id-overrides this means three LDAP searches.
+
+If the id-override data is downloaded once into the cache this can be handled
+more efficiently because the id-override data is already present on the local
+client and can be kept up-to-data in the similar scheme as sudo rules. For
```
ikerexxe, line 65: Do you mean in the same scheme as sudo rules or in **a** similar scheme as the sudo rules?

src/design-pages/idoverride_redesign.rst
```diff
@@ -0,0 +1,400 @@
...
+Solution overview
+*****************
+
+Implementation details
+**********************
+
+
+Bulk-loading id-overrides into the cache
+========================================
+While the id-override objects can be updates at runtime individually (although
```
ikerexxe, line 98: s/updates/updated

src/design-pages/idoverride_redesign.rst
```diff
@@ -0,0 +1,400 @@
...
+possible.
+
+The next steps, which might be done in parallel is adding the linking
+functionality to the common backend code and removing the id-override specific
+code paths in the user and group lookup code from the IPA provider. Updating
+the sss_override utility and integrating it in sssctl looks like another
+task which can be handle separately.
+
+Finally it should be checked if the cache request code can already handle the
+new setup of if some modifications are needed here as well.
```
ikerexxe, line 395: s/of/or

ikerexxe (changes requested):
I have added some concerns that need to be addressed before the change can be approved.

pbrezina:
I read the design again and I have some concerns. I don't think we can really rely on period updates of overrides instead of using direct look ups as we do now.

* the id-overrides have to be fetched before any user or group resolution, so SSSD can not return users and groups until this is done on first start up
* if id-override is created or changed, we would be returning incorrect user/group data until this change is picked up by periodic refresh
* potentially large amount of unused data will be stored on each client (each client does not have to use all id-overrides)

As far as I know, there are no performance issues with the current queries, but we would create easy to hit race condition (since updates will not be frequent).

With large data, we could also introduce performance issues due to the slow indices like the two bugzillas for autofs and sudo. But I believe this is a regression in ldb and Tomas is currently looking into it.

pbrezina:
I have identified these problems described in the design:

* **Local overrides** - local override gets deleted when cache object is removed even if it stays in the server
* Special handling of **Default Trust View**
* It is not allowed to have the same certificate in multiple overrides (same certificate for multiple users)
* `auto_provide_groups` and overrides definition/consistency... although I believe I have already fixed this in the past, we need to re-evaluate

@sumit-bose Is this a complete list? What do you think about https://github.com/SSSD/sssd.io/pull/59#issuecomment-1240602400?

There is also "Extending ID-mapping for AD with AD’s POSIX attributes (long term)" can you elaborate on the solution more?

pbrezina:
Bump @sumit-bose

src/design-pages/idoverride_redesign.rst
```diff
@@ -0,0 +1,400 @@
...
+Searches for user id-overrides are typically done by name, UID or certificate
+while group id-overrides are search by name or GID. A typical search filter
+combines the objectClass with the searched attribute, e.g.
+'(&(objectClass=override)(searchAttrName=searchValue))'. As seen above
+objectClass should not be used and in SSSD there is already the un-indexed
+objectCategory attribute which can be used to classify the object.
+
+For user id-overrides it is expected that the name and the UID are unique, the
+same is true for group id-overrides with respect to the GID. As seen above the
+attribute for the primary GID of the user does not have to be unique and if is
```
aplopez, line 218: s/if/it/

src/design-pages/idoverride_redesign.rst
```diff
@@ -0,0 +1,400 @@
...
+multiple accounts of a single physical user for different purposes. So we
+cannot avoid indexes with multiple entries in this case. However it is
+expected that for the above use-cases the multiple use of a single certificate
+is limited to small numbers in the order of 10 where adding new entries to the
+index should not cause a considerable slowdown. Nevertheless it might be worth
+to mention this in the documentation/man page.
+
+An potential additional optimization might be to use indexed attributes with
+names with a dedicated prefix, e.g. 'ido' for id-override. With this a search
+filter for an user id-override with a given UID can be just 'idoUidNumber=UID'
```
aplopez, line 237: s/an/a/

src/design-pages/idoverride_redesign.rst
```diff
@@ -0,0 +1,400 @@
...
+and the object can be directly returned from the indexed search without
+evaluating additional attributes. But since the additional evaluation would
+just be a string comparison I would expect only a limited speedup, if any.
+
+Managing id-overrides in SSSD's cache
+=====================================
+Currently there are two types of sources for the id-override object. They are
+either added locally with the 'sss_override' command or read from LDAP. With
+respect to LDAP currently only IPA is supported properly, but with this
+re-design adding support in plain LDAP or AD should be more easy.
```
aplopez, line 247: s/more easy/easier/

src/design-pages/idoverride_redesign.rst
```diff
@@ -0,0 +1,400 @@
...
+
+This separation also allows to keep the id-override in the cache while the
+original object is deleted from the cache. The primary purpose here is to
+avoid to loose the id-override if the original object was removed from the
+cache due to an error in SSSD, on the server side or in the configuration. The
+drawback is that the id-override becomes useless if the original object was
+removed on the serve side. But this issue is imo easier to handle than an
+id-override that was lost accidentally.
+
+The 'sss_override' command currently writes the id-override objects directly
```
aplopez, line 265: Maybe `sss_override` could be modified to identify, list and delete dead overrides.

src/design-pages/idoverride_redesign.rst
```diff
@@ -0,0 +1,400 @@
...
+id-override data from the server side and write it to the cache as fast as
+possible.
+
+The next steps, which might be done in parallel is adding the linking
+functionality to the common backend code and removing the id-override specific
+code paths in the user and group lookup code from the IPA provider. Updating
+the sss_override utility and integrating it in sssctl looks like another
+task which can be handle separately.
+
+Finally it should be checked if the cache request code can already handle the
```
aplopez, line 394: s/handle/handled/

PR #63, opened 2022-11-14
correct grammar in quickstart

### LINKED ISSUE #62: Incorrect grammar
https://github.com/SSSD/sssd.io/blob/master/src/docs/quick-start.rst

As your intent is likely for people to use the quickstart guide as a template to start their own sssd installation, you should change the **i.e.**  to  **e.g.**.

i.e. would imply that there is only one possible setting. And as the end user does not own the domain specified (sssd.io) that would contradict the documentation.

e.g. provides an example which is following in the worked example below

iandstanley:
https://github.com/SSSD/sssd.io/blob/master/src/docs/quick-start.rst

As your intent is likely for people to use the quickstart guide as a template to start their own sssd installation, you should change the **i.e.**  to  **e.g.**.

i.e. would imply that there is only one possible setting. And as the end user does not own the domain specified (sssd.io) that would contradict the documentation.

e.g. provides an example which is following in the worked example below

Fixes #62

### DIFF
--- a/src/docs/quick-start.rst
+++ b/src/docs/quick-start.rst
@@ -8,10 +8,10 @@ Quick Start IPA
 
 Before starting, make sure you have the following information.
 
-- Administrator credentials *i.e. admin*
-- FreeIPA domain name *i.e. sssd.io*
-- FreeIPA server hostname *i.e. ipa1.sssd.io*
-- FreeIPA server IP *i.e. 1.2.3.4*
+- Administrator credentials *e.g. admin*
+- FreeIPA domain name *e.g. sssd.io*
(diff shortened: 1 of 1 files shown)


pbrezina:
Thank you for your contribution, I agree. Can you also change it in AD and LDAP sections please?

iandstanley:
> Thank you for your contribution, I agree. Can you also change it in AD and LDAP sections please?

done

pbrezina:
Thank you, would you mind squashing it into single commit?

iandstanley:
Sorry I have limited access to github from my work machine - company bans most uses of it and I only have access via the web gui on my work machine. I think a merge can

PR #65, opened 2023-01-03
design pages: passkey Kerberos integration

ikerexxe:


src/design-pages/passkey_kerberos.rst
```diff
@@ -0,0 +1,223 @@
...
+
+Related ticket(s):
+------------------
+ * https://github.com/SSSD/sssd/issues/6228
+
+Problem statement
+-----------------
+This design is an extension of the "enable passkey authentication", where a
+centrally managed user can authenticate locally in a system with a passkey
+token. This might not be enough for SSSD in some environments like big
```
abbra, line 12: Please change this sentence to simply state that this design describes Kerberos pre-authentication mechanism using passkey authentication. Once Kerberos pre-authentication complete, a received initial ticket granting ticket (TGT) can be used for signle sign-on into other Kerberos-enabled services.

src/design-pages/passkey_kerberos.rst
```diff
@@ -0,0 +1,223 @@
...
+Solution overview
+-----------------
+The objective is to locally authenticate and issue a Kerberos ticket to a
+centrally managed user. For that purpose SSSD will need to be modified to
+accommodate the new workflow, while still allowing the old one, where the user
+was authenticated locally without issuing any Kerberos ticket. The new workflow
+will be used for online authentication against an IPA server. The old workflow
+will be used for any offline authentication and for AD and other LDAP servers.
+
+SSSD will send a pre_auth message to the Kerberos client. It will send an
```
abbra, line 43: ... "SSSD provides a pre-authentication method implementation (pre-auth plugin) to MIT Kerberos library. When both the Kerberos client and the KDC support the same pre-authentication method, KDC will respond to an AS-REQ message with a request to provide an assertion data required by a pre-authentication method. SSSD's pre-auth plugin will obtain the assertion data from the passkey. This information will be sent by the Kebreros client back to the KDC server as a part of the AS-REQ exchange."

src/design-pages/passkey_kerberos.rst
```diff
@@ -0,0 +1,223 @@
...
+  :alt: Sequence diagram
+
+* The user will try to login with a centrally managed account in the system.
+
+* PAM will relay this request to SSSD’s PAM responder.
+
+* The PAM responder will retrieve the user information from the cache, which
+  will provide this information including the authentication protocol (PASSKEY).
+
+* The PAM responder will try to pre_authenticate the user with PASSKEY.
```
abbra, line 79: pre-authenticate

src/design-pages/passkey_kerberos.rst
```diff
@@ -0,0 +1,223 @@
...
+
+* The IPA provider will execute the krb5_child, which in turn will communicate
+  with the Kerberos client.
+
+* Kerberos client will send an AS-REQ with the PASSKEY pre-authentication data.
+
```
abbra, line 85: Kerberos KDC will receive an AS-REQ and will query Kerberos database driver (KDB) for the information about the Kerberos principal. It will find out that the principal object supports pass key pre-authentication.

src/design-pages/passkey_kerberos.rst
```diff
@@ -0,0 +1,223 @@
...
+* Kerberos KDC will send an Access-Request with the username to ipa-otpd.
```
abbra, line 86: KDC side of the pre-authentication method implementation will use RADIUS protocol to communicate with IPA server side (`ipa-otpd` daemon listens over UNIX domain socket used by the KDC). The RADIUS message `Access-Request` will be sent to `ipa-otpd`.

src/design-pages/passkey_kerberos.rst
```diff
@@ -0,0 +1,223 @@
...
+
+* ipa-otpd will retrieve the user information and generate the assertion
+  request data (more information on the format in
+  :ref:`kerberos-kdc-ipa-otpd`), which will be used to fill the Reply-Message
+  that is returned to the Kerberos KDC.
+
+* Kerberos KDC returns an error with the list of authentication methods, which
```
abbra, line 93: The passkey assertion request data produced by the KDC side of the pre-authentication method implementation is returned to Kerberos client in an error response. This flow is common to all Kerberos protocol pre-authentication mechanisms.

src/design-pages/passkey_kerberos.rst
```diff
@@ -0,0 +1,223 @@
...
+  will include PASSKEY, and the assertion request data.
+
+* This information will be passed between the various components until it
```
abbra, line 96: I don't think this is what happens here. Please cross-verify with @pbrezina but at least I don't see this logic at all in the client side of the pre-auth mechanism. As krb5 pre-auth mechanism should work in a normal Kerberos application (like `kinit` tool), it definitely does not talk to PAM responder.

abbra (changes requested):
Hi, thank you for this document. I have few comments inline.

src/design-pages/passkey_kerberos.rst
ikerexxe: Changed

src/design-pages/passkey_kerberos.rst
ikerexxe: Updated

src/design-pages/passkey_kerberos.rst
ikerexxe: Fixed

src/design-pages/passkey_kerberos.rst
ikerexxe: Fixed

src/design-pages/passkey_kerberos.rst
ikerexxe: Fixed

src/design-pages/passkey_kerberos.rst
ikerexxe: Fixed

src/design-pages/passkey_kerberos.rst
```diff
@@ -0,0 +1,230 @@
...
+  :alt: Sequence diagram
+
+* The user will try to login with a centrally managed account in the system.
+
+* PAM will relay this request to SSSD’s PAM responder.
+
+* The PAM responder will retrieve the user information from the cache, which
+  will provide this information including the authentication protocol (PASSKEY).
+
+* The PAM responder will try to pre-authenticate the user with PASSKEY.
```
pbrezina, line 79: PAM at this point is not aware that passkey will be used. It is on KDC to decide which preauth methods are available, krb5 child then responds to questions generated by libkrb5. See `sss_krb5_responder()`.

See https://docs.google.com/document/d/1T6B4PyFq7uSZ7aDF-DH9UQG2H2fr5gBTe1JCyBxp3jo/edit?usp=sharing for more info on how the plugins work.

src/design-pages/passkey_kerberos.rst
pbrezina: Yes and no. It depends on who the caller is. If it is kinit, then SSSD is not involved (however, this is currently not supported, it is todo). If it is pam_sss then it is true, altough I would rather see that the communication goes from pam_sss <=> ipa-otpd, in this design it seems that the endpoint is the PAM responder which is not true.

src/design-pages/passkey_kerberos.rst
```diff
@@ -0,0 +1,230 @@
...
+
+* Authenticator data in b64 (String)
+
+* Assertion signature in b64 (String)
+
+.. _kerberos-kdc-ipa-otpd:
+
+Kerberos KDC - ipa-otpd
++++++++++++++++++++++++
+The Kerberos KDC is expected to provide the following json-formatted string
```
pbrezina, line 180: The messages are different, see description in https://github.com/SSSD/sssd/pull/6462

pbrezina (changes requested):
I would like to use this design page as a reference for registration of the new pa data number. Please, try to write it in more detail so it can be consumed by reader who is not familiar with fido2, sssd and ipa.

pbrezina:
I just submitted the request to IANA with this document: https://docs.google.com/document/d/1cFcEdm1dv6a0NWAoL3zJKH2Bpc9D3TxZS_QGEAJn-2k/edit?usp=sharing

src/design-pages/passkey_kerberos.rst
```diff
@@ -0,0 +1,230 @@
...
+to provide an assertion data required by a pre-authentication method. SSSD's
+pre-auth plugin will obtain the assertion data from the passkey. This
+information will be sent by libkrb5 back to the KDC server as a part of the
+AS-REQ exchange. This in turn will relay the request to the ipa-otpd, which
+will retrieve the user information and ask the passkey_child to validate the
+assertion. Once validated the Kerberos KDC will issue a ticket and send it to
+SSSD via libkrb5.
+
+.. image:: ../images/passkey_kerberos_1.png
+  :alt: Architecture diagram
```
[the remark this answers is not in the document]
ikerexxe: I'm using plantuml to draw all the diagrams as this is the recommended tool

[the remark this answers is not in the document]
pbrezina: Yes, but mermaid can be made part of the text so it can be easily updated and changes can be tracked.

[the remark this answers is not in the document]
pbrezina: We can also add https://github.com/sphinx-contrib/plantuml (or other extension) to generate the images, I have no problem with that. My goal is to move it from image to text.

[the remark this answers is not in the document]
abbra: We do use plantuml in FreeIPA as well. We also embed it in the text.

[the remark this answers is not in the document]
ikerexxe: Having the text instead of the image sounds perfect to me. If you can enable some script for plantuml that would be nice, as that's the tool that we should be using.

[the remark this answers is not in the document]
pbrezina: https://github.com/SSSD/sssd.io/pull/66

src/design-pages/passkey_kerberos.rst
```diff
@@ -0,0 +1,230 @@
...
+
+.. image:: ../images/passkey_kerberos_2.png
+  :alt: Sequence diagram
+
+* The user will try to login with a centrally managed account in the system.
+
+* PAM will relay this request to SSSD’s PAM responder.
+
+* The PAM responder will retrieve the user information from the cache, which
+  will provide this information including the authentication protocol (PASSKEY).
```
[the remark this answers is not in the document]
ikerexxe: I'd like to simplify the sequence diagram to avoid having a very big diagram that nobody understands. That's why I assumed that the cache already contained the user attributes, including the passkey mapping data. Does that make sense?

src/design-pages/passkey_kerberos.rst
```diff
@@ -0,0 +1,230 @@
...
+authentication. Once it is completed, the received initial ticket granting
+ticket (TGT) can be used for single sign-on into other Kerberos-enabled
+services.
+
+For the purpose of this feature, passkey is a FIDO2 compatible device supported
+by the libfido2 library.
+
+Use cases
+---------
+As a user, not only do I want to walk to any system on a network and log in
```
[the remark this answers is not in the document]
ikerexxe: Done

src/design-pages/passkey_kerberos.rst
```diff
@@ -0,0 +1,230 @@
...
+(desktop and/or terminal) using a passkey device connected to the system, but I
+also want to get a Kerberos ticket to identify myself to other services.
+
+Prerequisites
+-------------
+To register a new
+`pre-authentication data identifier <https://www.iana.org/assignments/kerberos-parameters/kerberos-parameters.xhtml#pre-authentication>`__
+in Kerberos for passkey it would be good to first finalize the IdP plugin
+design and security analysis requested by this community.
```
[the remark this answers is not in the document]
ikerexxe: Removed

src/design-pages/passkey_kerberos.rst
```diff
@@ -0,0 +1,230 @@
...
+
+Solution overview
+-----------------
+The objective is to locally authenticate and issue a Kerberos ticket to a
+centrally managed user. For that purpose SSSD will need to be modified to
+accommodate the new workflow, while still allowing the old one, where the user
+was authenticated locally without issuing any Kerberos ticket. The new workflow
+will be used for online authentication against an IPA server. The old workflow
+will be used for any offline authentication and for AD and other LDAP servers.
```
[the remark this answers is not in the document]
ikerexxe: Moved to the problem statement and rewritten part of it

src/design-pages/passkey_kerberos.rst
```diff
@@ -0,0 +1,230 @@
+Passkey authentication Kerberos integration
+=======================================================
+
+Related ticket(s):
+------------------
+ * https://github.com/SSSD/sssd/issues/6228
+
+Problem statement
```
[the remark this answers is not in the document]
ikerexxe: I've done a general revision of this section by including new information to answer the questions that you provided and rewritten the existing text to make sense of it

src/design-pages/passkey_kerberos.rst
```diff
@@ -0,0 +1,230 @@
...
+SSSD provides a pre-authentication method implementation (pre-auth plugin) to
+MIT Kerberos library. When both the libkrb5 and the KDC support the same
+pre-authentication method, KDC will respond to an AS-REQ message with a request
+to provide an assertion data required by a pre-authentication method. SSSD's
+pre-auth plugin will obtain the assertion data from the passkey. This
+information will be sent by libkrb5 back to the KDC server as a part of the
+AS-REQ exchange. This in turn will relay the request to the ipa-otpd, which
+will retrieve the user information and ask the passkey_child to validate the
+assertion. Once validated the Kerberos KDC will issue a ticket and send it to
+SSSD via libkrb5.
```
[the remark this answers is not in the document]
ikerexxe: Done, but I don't know if this will make more sense

src/design-pages/passkey_kerberos.rst
```diff
@@ -0,0 +1,230 @@
...
+.. image:: ../images/passkey_kerberos_1.png
+  :alt: Architecture diagram
+
+Implementation details
+----------------------
+
+General overview
+****************
+When SSSD support for passkey authentication is enabled, libkrb5 would
+advertise a support for passkey during AS-REQ request. If information about the
```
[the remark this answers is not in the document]
ikerexxe: Changed to `KDC advertises`...

src/design-pages/passkey_kerberos.rst
```diff
@@ -0,0 +1,230 @@
...
+Kerberos principal on the KDC side allows use of passkey authentication, KDC
+would respond with a pre-authentication data to choose passkey authentication.
+
+The following sequence diagram illustrates the workflow taking into account all
+the components involved in it:
+
+.. image:: ../images/passkey_kerberos_2.png
+  :alt: Sequence diagram
```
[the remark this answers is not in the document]
ikerexxe: Resolving as we are already discussing it in https://github.com/SSSD/sssd.io/pull/65#discussion_r1129420133

src/design-pages/passkey_kerberos.rst
ikerexxe: Updated

src/design-pages/passkey_kerberos.rst
ikerexxe: I think this is fixed, if not, please reopen

src/design-pages/passkey_kerberos.rst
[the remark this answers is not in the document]
ikerexxe: Understood and fixed. Please check the new version

src/design-pages/passkey_kerberos.rst
ikerexxe: Fixed

ikerexxe:
I forgot to mention that this is ready for review again.

pbrezina (approved):
Thank you, ack.

src/design-pages/passkey_kerberos.rst
```diff
@@ -0,0 +1,262 @@
...
+  libkrb5.
+
+Authentication method policy
+****************************
+
+This design adds some complexity to the passkey authentication, and this may
+lead to a degraded user experience. For this reason, the following
+authentication policy is implemented:
+
+* In the IPA case, if the server is online and a passkey entry is found in the
```
sumit-bose, line 70: Hi,

I think in the online case a "policy" "try only methods which will return a Kerberos ticket" makes sense. But even "if online and a passkey entry is found" this might not be possible wither because the IPA server does not has the Kerberos plugin installed or the user entry does not have `passkey` listed in the user-auth-types. In those cases I wonder if having passkey authentication available by default for offline authentication might irritate the users?

So it might be worth to store the authentication methods which are available online in the user object to have a consistent experience when doing offline authentication?

Additionally, to make the passkey usage more flexible, it might be worth to add an option for the backends to allow passkey authentication although not supported via Kerberos on the server side. This might be useful for the AD use-case as well to allow passkey authentication for users which have the passkey data set in the user entry. Since this would be an exception to the  "try only methods which will return a Kerberos ticket" "policy" this option should not be enable by default. But in environments which Kerberos tickets are not that important it would help to support the usage of local passkey authentication. If the passkey prompt should come first in those cases and can be skipped if users wants to have a login method which returns a Kerberos ticket or if it should come last after method which can get a Kerberos ticket might then be something which should be configurable in the PAM responder to have a consistent behavior if multiple backends are configured.

For `auth_provider = ldap` this option can be enabled by default, i.e. if the user has passkey data configured local passkey authentication is used and if not authentication against the LDAP server with an authenticated LDAP bind is used.

I wonder if this option can be boolean of if there might be other use-cases like not allowing passkey when online but only when offline? Or only allowing passkey authentication (this would look like an `auth_provider = passkey` but since we support Kerberos with passkey in IPA I think it would be better to have this as an option for all the existing providers).

About skipping passkey authentication. I think it would be good to allow this either if the user enters something at the interactive prompt or nothing at the pin prompt. This would cover the case where a pin is not needed but the interactive prompt might be extended with  e.g. "... or press 's' and then enter to skip passkey authentication".

bye,
Sumit

abbra: given that this policy should apply to all passwordless methods, including smartcards and IdP (in IPA case), may be it would be best to have a bit less specific set of options? 

E.g. something like  `local_auth_policy = match | disable | enable` where `match` is to match offline and online state for Kerberos methods, `disable` would disallow those methods that do not lead to Kerberos ticket, and `enable` would allow any local methods.

sumit-bose: Hi,

about `passkey_auth_priority`, I think this functionality should be part of a more generic `order` option for the ´[prompting]` section. If you set `passkey` first with this option, e.g. `order = passkey, 2fa, password`, then passkey authentication would be tried first if available for the user.

About `allow_local_passkey`, I agree with Alexander that this option should be more generic to cover other authentication methods as well. However, I think it would be good to have a way to only enable/allow specific method, so maybe options like `enable:passkey` to only allow passkey for local authentication. This can be handled with the `order` option as well, e.g. `order = passkey, password` would only allow passkey authentication and no other local methods like e.g. Smartcard. But since this will control the behavior for all configured domains it mightr be better to have something on the domain level as well. E.g. if there are two domains configured and one domain should allow local passkey authentication while the other domains local Smartcard authentication.

Additionally, a fourth option `only`  for ` local_auth_policy` which would ignore the online methods and only offer the local ones might be useful. This would cover e.g. the AD use case where only passkey authentication should be used.

bye,
Sumit

sumit-bose:
Hi,

thanks for the updates. After thinking more about the options I'd like to make the following suggestions.

About `local_auth_policy`, first I would like to suggest describe what this option would mean for other types of authentication providers as well, i.e. LDAP and proxy provider. Currently those two provider only support a password base authentication in the backend so the  default policy here might be somehting like "Only do what the backend/provider can". But e.g with respect to Smartcard authentication of local user with the help of the proxy provider a configurable fallback to local authentication would be useful. In this context I wonder if it would make sense to handle `local_auth_policy` also in the context of `auth_provider = none` so that Smartcard authentication for local users can be enabled by setting `auth_provider = none` and `local_auth_policy = only`. Imo this seems to be more logical than having `auth_provider = proxy` and some reserved module name as mentioned in https://github.com/SSSD/sssd/pull/6633/files#diff-0431aed9691c8b963c60df60bfa4f49d3d5f1ce202749481c7fcc7b2ac617eeeR144 . (Btw, although I guess @alexey-tikhonov would not like to hear it but adding passkey authentication for local users can be done the same way with a `[passkey/domain_name/user_name]` section and an option which contains the passkey key id and public key strings as e.g. stored in LDAP). @spoore1, do you have any ideas how a good configuration for Smartcard authentication for local users should look like?

My second point about `local_auth_policy` is that I currently do not see the difference between `match` and `disable`. E.g. for Kerberos based authentication both would mean to allow only methods which would get a Kerberos ticket in online and offline case, if I understand it correctly.

About `auth_order`, I think it would offer greater flexibility to handle this in the `[prompting/...]` section. E.g. it would allow to offer different methods for different PAM services. So that in the AD case you can configure that you want Smartcard authentication first as most secure method to get a Kerberos ticket for general authentication but use passkey as first method for simplicity for "su" and "sudo" where Kerberos tickets are not needed. This might be specified in a section like `[prompting/order/sudo]`.

Additionally, with moving `auth_order` out of the domain section some ambiguities or dependencies with `local_auth_policy` might be resolved. E.g. what would it mean if `auth_order` contains `passkey` but `local_auth_policy` does not allow it? Or the other way round, `local_auth_policy = enable:passkey` but passkey is not listed in `auth_order`? By moving it into the `[prompting/...] section that tasks are move clear imo, the backend tells what is allowed and the `[prompting/..]` options control what is shown to the user.

bye,
Sumit

ikerexxe:
> (Btw, although I guess @alexey-tikhonov would not like to hear it but adding passkey authentication for local users can be done the same way with a `[passkey/domain_name/user_name]` section and an option which contains the passkey key id and public key strings as e.g. stored in LDAP).

I'm not against supporting local users (in the future), but I clearly don't like this solution. We shouldn't "pollute" the `/etc` folder with a database. This folder is meant to be used for configuration files, and by adding the mapping data we are changing its purpose. This is specially meaningful for immutable distributions (Fedora CoreOS, Fedora Silverblue, etc.).

abbra:
While Silverblue is immutable, `/etc` is not mounted read-only. In this sense we probably aren't too different from certmapping configuration using a full certificate blob, as setting up a passkey in the configuration would mean the same, effectively.

Another solution is to revive local overrides and store in the actual database, like it was designed for in past. We had few use cases to store LDAP attributes in the local overrides but none really got consumed by expected external apps:
- GNOME wanted to use rich metadata for user photo/icon
- systemd wanted some data for querying related data (https://github.com/SSSD/sssd/issues/5104)

alexey-tikhonov:
> so that Smartcard authentication for local users can be enabled by setting `auth_provider = none` and `local_auth_policy = only`. Imo this seems to be more logical than having `auth_provider = proxy` and some reserved module name

Sounds interesting.


> (Btw, although I guess @alexey-tikhonov would not like to hear it but adding passkey authentication for local users can be done the same way with a `[passkey/domain_name/user_name]` section and an option which contains the passkey key id and public key strings as e.g. stored in LDAP).

If this won't be different as compared with "smart card auth for local users" then technically this makes sense.
Motivation to support this use case is a different question.


> certmapping configuration using a full certificate blob

Do we ever support mapping by full blob for local users case?
Default 'maprule' is "The RULE_NAME for the “files” provider which tries to find a user with the same name."

spoore1:
> 
>  @spoore1, do you have any ideas how a good configuration for Smartcard authentication for local users should look like?
> 

I'm not sure I have a "good" configuration :-)

In most of my local smart card tests i tend to use a very basic config with `id_provider = files`.  I know this needs to change soon.

In addition in most of my testing, I specify a `[certmap/domain_name/user_name]` entry to match the certificate(s) on the card.

This is a very basic example of one that I used recently:

```
[sssd]
services = nss, pam
domains = shadowutils

[nss]

[pam]
pam_cert_auth = True

[domain/shadowutils]
id_provider = files

[certmap/shadowutils/localuser1]
matchrule = <SUBJECT>.*CN=localuser1*
```

I have not yet tried using a full certificate blob in certmap in sssd.conf yet.   Is that even possible today?   Or would that require something like a local ldap auth_provider to store/map the certificates to users?

alexey-tikhonov:
> I have not yet tried using a full certificate blob in certmap in sssd.conf yet. Is that even possible today? Or would that require something like a local ldap auth_provider to store/map the certificates to users?

I would guess it requires `sss_override user-add $user --certificate` - i.e. local override.
But Alexander mentioned this as "Another solution".

sumit-bose:
> > @spoore1, do you have any ideas how a good configuration for Smartcard authentication for local users should look like?
> 
> @sumit-bose
> 
> I'm not sure I have a "good" configuration :-)
> 
> In most of my local smart card tests i tend to use a very basic config with `id_provider = files`. I know this needs to change soon.
> 
> In addition in most of my testing, I specify a `[certmap/domain_name/user_name]` entry to match the certificate(s) on the card.
> 
> This is a very basic example of one that I used recently:
> 
> ```
> [sssd]
> services = nss, pam
> domains = shadowutils
> 
> [nss]
> 
> [pam]
> pam_cert_auth = True
> 
> [domain/shadowutils]
> id_provider = files
> 
> [certmap/shadowutils/localuser1]
> matchrule = <SUBJECT>.*CN=localuser1*
> ```

Hi,

I meant how do you think the new proxy-provider based configuration should look like:
```
[domain/shadowutils]
id_provider = proxy
proxy_lib_name = files
```
has to be used for a start. The question is now how to tell not to do proxy authentication but local Smartcard (or passkey) authentication. Given that we are talking about the new `local_auth_policy` parameter I think
```
auth_provider = none
local_auth_policy = only
```
would be most straight forward and would enable all local methods the PAM responder is configured to handle. An alternative would be something like
```
auth_provider = proxy
proxy_pam_target = sss_smartcard
```
where `sss_smartcard` is a special keyword which does not correspond to a PAM config file but is evaluated internally by the proxy provider.

Which version would you prefer or which do you think is most easy to understand for users and customers?

bye,
Sumit

> 
> I have not yet tried using a full certificate blob in certmap in sssd.conf yet. Is that even possible today? Or would that require something like a local ldap auth_provider to store/map the certificates to users?

sumit-bose:
> > I have not yet tried using a full certificate blob in certmap in sssd.conf yet. Is that even possible today? Or would that require something like a local ldap auth_provider to store/map the certificates to users?
> 
> I would guess it requires `sss_override user-add $user --certificate` - i.e. local override. But Alexander mentioned this as "Another solution".

Hi,

yes, that what the local certificate override is for.

@alexey-tikhonov, do you have any opinions on the `local_auth_policy` option or how to configure the prompting order?

bye,
Sumit

spoore1:
> Hi,
> 
> I meant how do you think the new proxy-provider based configuration should look like:
> 
> ```
> [domain/shadowutils]
> id_provider = proxy
> proxy_lib_name = files
> ```
> 
> has to be used for a start. The question is now how to tell not to do proxy authentication but local Smartcard (or passkey) authentication. Given that we are talking about the new `local_auth_policy` parameter I think
> 
> ```
> auth_provider = none
> local_auth_policy = only
> ```
> 
> would be most straight forward and would enable all local methods the PAM responder is configured to handle. An alternative would be something like
> 
> ```
> auth_provider = proxy
> proxy_pam_target = sss_smartcard
> ```
> 
> where `sss_smartcard` is a special keyword which does not correspond to a PAM config file but is evaluated internally by the proxy provider.
> 
> Which version would you prefer or which do you think is most easy to understand for users and customers?
> 
> bye, Sumit

Honestly, I'm not sure.  

At first glance my preference would be option 1 (`auth_provider = none`).  Seems straight forward and to the point:  There is no external authentication provider and only local methods should be tried.

But, the advantage of option 2 (`auth_provider = proxy`) could be that you could technically don't need the `auth_provider` line because it defaults to `id_provider`'s setting of `proxy` right?

Is there a reason that `local_auth_policy` couldn't work with `auth_provider = proxy` the same as described with `none`?    I'm guessing that would further complicate the proxy code even more than using `proxy_pam_target`?

If so, then I think I'd vote for option 1.

alexey-tikhonov:
Imo,
```
auth_provider = none
local_auth_policy = only
```
used together looks good (`auth_provider = none` alone feels a bit like "auth disabled").

alexey-tikhonov:
> the advantage of option 2 (`auth_provider = proxy`) could be that you could technically don't need the `auth_provider` line

But why care? IIUC, authselect will take care of is (right, @pbrezina?)

spoore1:
> > Honestly, I'm not sure.
> > At first glance my preference would be option 1 (`auth_provider = none`). Seems straight forward and to the point: There is no external authentication provider and only local methods should be tried.
> > But, the advantage of option 2 (`auth_provider = proxy`) could be that you could technically don't need the `auth_provider` line because it defaults to `id_provider`'s setting of `proxy` right?
> 
> Hi,
> 
> yes, the `id_provider` value is inherited.
> 
> > Is there a reason that `local_auth_policy` couldn't work with `auth_provider = proxy` the same as described with `none`? I'm guessing that would further complicate the proxy code even more than using `proxy_pam_target`?
> 
> No, imo `local_auth_policy` should work with `auth_provider = proxy` as well the same way as with the other providers. The question is more related to `auth_provider = none` which currently is just an option to load no provider at all. So if you think as well that
> 
> ```
> [domain/shadowutils]
> id_provider = proxy
> proxy_lib_name = files
> auth_provider = none
> local_auth_policy = only
> ```
> 
> is a reasonable way to enable Smartcard (or other local) authentication for local users we can keep this and make `local_auth_policy` an option for all providers including `none`.

I'm good with that.   +1

sumit-bose:
> > the advantage of option 2 (`auth_provider = proxy`) could be that you could technically don't need the `auth_provider` line
> 
> But why care? IIUC, authselect will take care of is (right, @pbrezina?)

Hi,

afaik authselect tries to avoid to touch sssd.conf. Maybe this is something which should be better handled with Ansible roles if we want to help user to not edit sssd.conf directly?

bye,
Sumit

src/design-pages/passkey_kerberos.rst
```diff
@@ -0,0 +1,309 @@
...
+++++
+
+The following domain option can be used to tune the authentication policy:
+
+.. _local-auth-policy:
+
+* local_auth_policy: local authentication methods policy. Some backends (i.e.
+  LDAP, proxy provider) only support a password base authentication, in such
+  cases authentication is only performed with the methods supported by the
+  backend.
```
sumit-bose, line 270: Hi,

I think in this paragraph some additional context would be beneficial. E.g.
```
Some backends (i.e. LDAP, proxy provider) only support a password base authentication while other can handle PKINIT based Smartcard authentication (AD, IPA), two-factor authentication (IPA) or other methods against a central instance. By default in such cases authentication is only performed with the methods supported by the backend. To allow more convenient or secure authentication methods which are supported by SSSD but not by the backend in cases where a central authentication is not strictly required the `local_auth_policy` option is added.
```

bye,
Sumit

ikerexxe: I included this paragraph. Thank you!

sumit-bose:
Hi,

thanks for the update, I'm fine with the current version. @abbra, any final comments?

bye,
Sumit

abbra:
Thank you, I don't have more comments.

ikerexxe:
In that case, @sumit-bose and @abbra can you approve the PR?

abbra (approved):

sumit-bose (approved):

PR #78, opened 2024-01-08
docs: add files provider removal document

justin-stephenson:
Include steps on how to switch to equivalent proxy provider configuration.

### DIFF
--- a/src/docs/files-provider-deprecation.rst
+++ b/src/docs/files-provider-deprecation.rst
@@ -1,0 +1,102 @@
+Files Provider Removal
+######################
+
+The SSSD team has announced the `removal of the files provider <https://pagure.io/fesco/issue/3107>`_ feature in recent versions of SSSD in Fedora/CentOS Stream/RHEL. This document describes how to switch to using the SSSD 'proxy' provider to replace files provider functionality.
+
+Practically, there are only two use cases that currently justify usage of the files provider feature:
+
+* smart card authentication of local users
+* session recording for local users
+
+For both cases the proxy provider is a viable substitute. Otherwise, there is no benefit in SSSD handling local users. The SSSD proxy provider is just a relay, an intermediary configuration. SSSD connects to its proxy service, and then that proxy loads the specified libraries. This allows SSSD to use some resources that it otherwise would not be able to use.
+
+Follow the steps below based on the use case you are using.
+
+Smartcard Authentication for Local Users
+****************************************
+
+An example of the old `files provider` smartcard configuration which will no longer function properly is:
+
+.. code-block:: console
+
+    [pam]
+    pam_cert_auth = True
+
+    [domain/shadowutils]
+    id_provider = files
+
+Now let's look at the steps to setup the proxy provider replacement. First, make sure to install the ``sssd-proxy`` package.
+
+.. code-block:: console
+
+    $ dnf install sssd-proxy
+
+Configure SSSD for smart card authentication as below:
+
+.. code-block:: console
+
+    [domain/myproxy]
+    id_provider = proxy
+    proxy_lib_name = files
+    local_auth_policy = only
+
+.. note:: An explicit **proxy_pam_target** is only needed if you want Smartcard authentication be an additional method besides the method provided by the **proxy_pam_target** (in this case you should also use **local_auth_policy = enable:smartcard**). This is typically not needed for password authentication because pam_unix should be present in the PAM configuration to handle this.
+
+.. note:: The domain name in the `certmap` rule needs to be updated if the domain name was modified from the previous configuration, **shadowutils** to **myproxy** as shown here.
+
+Choose one of the following authselect profiles and run the appropriate command:
+
+.. code-block:: bash
+
+    authselect select sssd with-smartcard
+    authselect select sssd with-smartcard-lock-on-removal
+    authselect select sssd with-smartcard-required
+
+Now you are done!
+
+Session Recording for Local Users
+*********************************
+
+Configure SSSD for session recording as below:
+
+.. code-block:: console
+
+    [sssd]
+    services=nss, pam
+    domains=nssfiles
+
+    [domain/nssfiles]
+    id_provider=proxy
+    proxy_lib_name=files
+    proxy_pam_target=sssd-shadowutils
+
+The ``proxy_lib_name`` option specifies which existing NSS library to proxy identity requests through.
+
+The ``proxy_pam_target`` specifies the target to which PAM must proxy as an authentication provider. This PAM target is a file containing PAM stack information in the default PAM directory, ``/etc/pam.d/``. The file `sssd-shadowutils <https://github.com/SSSD/sssd/blob/master/src/examples/sssd-shadowutils>`_ is packaged with upstream SSSD and shipped in Fedora.
+
+.. note:: Ensure that the proxy PAM target stack does not recursively include pam_sss.so.
+
+.. code-block:: console
+
+    ~# cat /etc/pam.d/sssd-shadowutils
+    #%PAM-1.0
+    auth        [success=done ignore=ignore default=die] pam_unix.so nullok try_first_pass
+    auth        required      pam_deny.so
+
+    account     required      pam_unix.so
+    account     required      pam_permit.so
+
+Add SSSD session recording configuration, refer to `man sssd-session-recording` for more details:
+
+.. code-block:: console
+
+        [session_recording]
+        scope = some
+        users = contractor1, contractor2
+        groups = students
+
+Next it is required to set the `sss` module as the first module in the list for the `passwd` and `group` databases of ``/etc/nsswitch.conf``. On systems managed by authselect, the following command should be run:
+
+.. code-block:: bash
+
+    authselect select sssd with-tlog
--- a/src/contents.rst
+++ b/src/contents.rst
@@ -55,6 +55,7 @@ Table of Contents
    docs/introduction
    Architecture <docs/architecture>
    docs/reporting-bugs
+   docs/files-provider-deprecation
 
 .. toctree::
    :caption: Active Directory


justin-stephenson:
@alexey-tikhonov This is ready for review when you have time. One thing I'm not sure about is if any additional steps should be documented here for `smart card authentication of local users` use case.

src/docs/files-provider-deprecation.rst
```diff
@@ -0,0 +1,60 @@
...
+######################
+
+The SSSD team has announced the `removal of the files provider <https://pagure.io/fesco/issue/3107>`_ feature in recent versions of SSSD in Fedora/CentOS Stream/RHEL. This document describes how to switch to using the SSSD 'proxy' provider to replace files provider functionality.
+
+Practically, there are only two use cases that currently justify usage of the files provider feature:
+
+* smart card authentication of local users
+* session recording for local users
+
+For both cases the proxy provider is a viable substitute. Otherwise, files provider functionality does not provide any benefit and should remain disabled.
```
alexey-tikhonov, line 11: Maybe better something like "Otherwise there are no benefits in SSSD handling local users"?

src/docs/files-provider-deprecation.rst
```diff
@@ -0,0 +1,60 @@
...
+
+    [domain/nssfiles]
+    id_provider=proxy
+    proxy_lib_name=files
+    proxy_pam_target=sssd-shadowutils
+
+The ``proxy_lib_name`` option specifies which existing NSS library to proxy identity requests through.
+
+The ``proxy_pam_target`` specifies the target to which PAM must proxy as an authentication provider. This PAM target is a file containing PAM stack information in the default PAM directory, ``/etc/pam.d/``. The following ``sssd-shadowutils`` file exists on RHEL-based Linux systems.
+
```
alexey-tikhonov, line 40: Maybe better reference available upstream https://github.com/SSSD/sssd/blob/master/src/examples/sssd-shadowutils ?

src/docs/files-provider-deprecation.rst
```diff
@@ -0,0 +1,60 @@
...
+    auth        [success=done ignore=ignore default=die] pam_unix.so nullok try_first_pass
+    auth        required      pam_deny.so
+
+    account     required      pam_unix.so
+    account     required      pam_permit.so
+
+nsswitch changes
+****************
+
+In the `session recording for local users` case it is required set the `sss` module as the first module in the list for the `passwd` and `group` databases of ``/etc/nsswitch.conf``. The following command should be run:
```
alexey-tikhonov, line 56: ... requires **to** set...?

src/docs/files-provider-deprecation.rst
```diff
@@ -0,0 +1,60 @@
...
+
+.. code-block:: bash
+
+    authselect select sssd with-tlog
```
alexey-tikhonov, line 60: Maybe "could be used on a systems, managed by authselect"?

Current text, imo, is a bit too Fedora-centric for upstream resource...

src/docs/files-provider-deprecation.rst
alexey-tikhonov: Upstream sssd.spec packages this file:
https://github.com/SSSD/sssd/blob/998503210b2644dda35091ce87531d3ee31a94b4/contrib/sssd.spec.in#L808
And Fedora ships it.

alexey-tikhonov:
> One thing I'm not sure about is if any additional steps should be documented here for `smart card authentication of local users` use case.

Imo, makes sense.

IIUC, something like
https://github.com/SSSD/sssd/blob/998503210b2644dda35091ce87531d3ee31a94b4/src/tests/intg/test_pam_responder.py#L63
is required in this case.

src/docs/files-provider-deprecation.rst
justin-stephenson: Fixed.

src/docs/files-provider-deprecation.rst
justin-stephenson: Updated.

src/docs/files-provider-deprecation.rst
justin-stephenson: Fixed.

justin-stephenson:
> > One thing I'm not sure about is if any additional steps should be documented here for `smart card authentication of local users` use case.
> 
> Imo, makes sense.
> 
> IIUC, something like https://github.com/SSSD/sssd/blob/998503210b2644dda35091ce87531d3ee31a94b4/src/tests/intg/test_pam_responder.py#L63 is required in this case.
> 
> (with a "normal" `proxy_lib_name`, of course)

Hi @sumit-bose can you clarify what should be documented here for `smart card authentication of local users` use case (switching from files provider -> proxy provider) ?

sumit-bose:
> > > One thing I'm not sure about is if any additional steps should be documented here for `smart card authentication of local users` use case.
> > 
> > 
> > Imo, makes sense.
> > IIUC, something like https://github.com/SSSD/sssd/blob/998503210b2644dda35091ce87531d3ee31a94b4/src/tests/intg/test_pam_responder.py#L63 is required in this case.
> > (with a "normal" `proxy_lib_name`, of course)
> 
> Hi @sumit-bose can you clarify what should be documented here for `smart card authentication of local users` use case (switching from files provider -> proxy provider) ?

Hi,

it depends, but since with the `files` provider only Smartcard authentication was possible something like

[domain/myproxy]
id_provider = proxy
proxy_lib_name = files
local_auth_policy = only

(like in https://github.com/SSSD/sssd/blob/998503210b2644dda35091ce87531d3ee31a94b4/src/tests/intg/test_pam_responder.py#L59) would be sufficient.

An explicit `proxy_pam_target` is only needed if you want Smartcard authentication be an additional method besides the method provided by the `proxy_pam_target`. Please note that this is typically not needed for password authentication because `pam_unix` should be present in the PAM configuration to handle this (the integration tests are using a PAM configuration which only contains `pam_sss`).

HTH

bye,
Sumit

justin-stephenson:
> > > > One thing I'm not sure about is if any additional steps should be documented here for `smart card authentication of local users` use case.
> > > 
> > > 
> > > Imo, makes sense.
> > > IIUC, something like https://github.com/SSSD/sssd/blob/998503210b2644dda35091ce87531d3ee31a94b4/src/tests/intg/test_pam_responder.py#L63 is required in this case.
> > > (with a "normal" `proxy_lib_name`, of course)
> > 
> > 
> > Hi @sumit-bose can you clarify what should be documented here for `smart card authentication of local users` use case (switching from files provider -> proxy provider) ?
> 
> Hi,
> 
> it depends, but since with the `files` provider only Smartcard authentication was possible something like
> 
> [domain/myproxy] id_provider = proxy proxy_lib_name = files local_auth_policy = only
> 
> (like in https://github.com/SSSD/sssd/blob/998503210b2644dda35091ce87531d3ee31a94b4/src/tests/intg/test_pam_responder.py#L59) would be sufficient.
> 
> An explicit `proxy_pam_target` is only needed if you want Smartcard authentication be an additional method besides the method provided by the `proxy_pam_target`. Please note that this is typically not needed for password authentication because `pam_unix` should be present in the PAM configuration to handle this (the integration tests are using a PAM configuration which only contains `pam_sss`).
> 
> HTH
> 
> bye, Sumit

Thank you, I added this information into the document.

justin-stephenson:
I revised the document a good amount to separate each use case (Smart card and session recording) then describing each the steps needed.

Please re-review.

alexey-tikhonov (approved):

alexey-tikhonov:
ACK

martinpitt:
Thanks @justin-stephenson @sumit-bose and @alexey-tikhonov ! I confirm that this works well, wrt. https://bugzilla.redhat.com/show_bug.cgi?id=2260445

src/docs/files-provider-deprecation.rst
```diff
@@ -0,0 +1,90 @@
...
+
+.. code-block:: console
+
+    [domain/myproxy]
+    id_provider = proxy
+    proxy_lib_name = files
+    local_auth_policy = only
+
+.. note:: An explicit **proxy_pam_target** is only needed if you want Smartcard authentication be an additional method besides the method provided by the **proxy_pam_target**. This is typically not needed for password authentication because pam_unix should be present in the PAM configuration to handle this.
+
```
sumit-bose, line 34: Hi.

in the case where `proxy_pam_target` is set you most probably want to use `local_auth_policy = enable:smartcard` because with `only` the authentication provided by `proxy_pam`target` would be ignored. So I would suggest to either mention this in the note as well or add a second example snippet. I guess I would prefer the second.

As a general comment, I wonder if it would make sense to add an example of a possible old configuration as well to make it easier for new admins to understand what they should look for. I'm thinking about cases where "I have inherited these systems from a person which is not in the company anymore and after the last updated nothing is working anymore ....". With this it would e.g.  for local Smartcard authentication more obvious to mention that the domain name should be kept or if not the domain name in the [certmap/...] sections have to be updated as well.

bye,
Sumit

justin-stephenson: > in the case where `proxy_pam_target` is set you most probably want to use `local_auth_policy = enable:smartcard` because with `only` the authentication provided by `proxy_pam`target` would be ignored. So I would suggest to either mention this in the note as well or add a second example snippet. I guess I would prefer the second.

I added it to the existing `note` line, to keep the document concise.
> 
> As a general comment, I wonder if it would make sense to add an example of a possible old configuration as well to make it easier for new admins to understand what they should look for. I'm thinking about cases where "I have inherited these systems from a person which is not in the company anymore and after the last updated nothing is working anymore ....". With this it would e.g. for local Smartcard authentication more obvious to mention that the domain name should be kept or if not the domain name in the [certmap/...] sections have to be updated as well.

Could you provide an example configuration? I'll add it here as you suggested.

justin-stephenson: I added a sample old/working smart card sssd.conf configuration.

sumit-bose (approved):
Hi,

the Smartcard section looks good, ACK.

bye,
Sumit

alexey-tikhonov (approved):

PR #79, opened 2024-01-17
design pages: Passwordless-GDM integration

ikerexxe:
Passwordless authentication from the GUI.

### DIFF
--- a/src/design-pages/passwordless_gdm.rst
+++ b/src/design-pages/passwordless_gdm.rst
@@ -1,0 +1,506 @@
+Passwordless-GDM authentication integration
+=======================================================
+
+Related ticket(s):
+------------------
+ * https://github.com/SSSD/sssd/issues/7069
+
+Problem statement
+-----------------
+Passwordless authentication is becoming increasingly popular. SSSD and FreeIPA
+already provide several authentication mechanisms which make use of it:
+`Smart Cards <https://sssd.io/design-pages/smartcards.html>`__,
+`External Identity Providers
+<https://freeipa.readthedocs.io/en/latest/designs/external-idp/external-idp.html>`__
+(EIdP) and `Passkeys <https://sssd.io/design-pages/passkey_kerberos.html>`__.
+Unfortunately, the integration of these mechanisms into the graphical interface
+leaves much to be desired. Some of them may work in a degraded mode, while
+others can’t be used at all.
+
+SSSD and the GUI should be better integrated to make all these authentication
+mechanisms effortless for the user. This would increase the overall security of
+the system, by providing the benefits of these authentication mechanisms, i.e.
+passwordless, MFA, etc.
+
+SSSD and `GDM <https://wiki.gnome.org/Projects/GDM>`__ are working together to
+provide a set of interfaces that can be used to enable these authentication
+mechanisms in Linux’s GUI. While the initial work targets SSSD-GDM integration,
+the objective is that these interfaces can be used by any other desktop
+environment.
+
+Use cases
+---------
+* As a centrally managed user, I want to choose the authentication mechanism
+  to login from the graphical interface so that I can select the one that best
+  suits my needs.
+
+* As a centrally managed user, I want to use an external identity provider
+  (IdP) to login from the graphical interface so that the user interface is
+  easy to use and consistent across all authentication mechanisms in the
+  distribution.
+
+* As a centrally managed user, I want to select the smart card identity to
+  login from the graphical interface so that the authentication is performed
+  with the correct credentials.
+
+* As a centrally managed user, I want to use a passkey to login from the
+  graphical interface so that the user interface is easy to use and consistent
+  across all authentication mechanisms in the distribution.
+
+* As a centrally managed user, I want to get notified when the passkey
+  or smartcard authentication has been performed locally so that I am aware
+  that the user experience might be affected.
+
+Solution overview
+-----------------
+The objective is to provide usable workflows for users to authenticate in the
+graphical interface using passwordless authentication mechanisms. To do this,
+we first need to design a communication protocol between SSSD and GDM.
+
+The APIs can be implemented in two ways, either by implementing them in
+`D-BUS <https://www.freedesktop.org/wiki/Software/dbus/>`__, or by extending
+the PAM conversation to include the new authentication mechanisms. There
+doesn’t seem to be any advantage in using D-BUS, whereas a PAM-level API would
+give other desktop environments the ability to offer similar features. For this
+reason, the GDM PAM extension has been selected.
+
+GDM already provides an interface in ``/usr/include/gdm/gdm-pam-extensions.h``
+that SSSD supports. Unfortunately extending this binary interface is hard, so
+a JSON based protocol is used. It’s easier to extend and manage, making it
+easier to adapt the different packages involved. The previous implementation
+will continue to be maintained for backward compatibility.
+
+As for the workflows, they have to provide the user with a way to interact
+(i.e. insert the hardware token or enter the PIN), while allowing communication
+with the device or the external provider. Each authentication mechanism needs
+its own sequence diagram definition, which will be explained later in this
+document.
+
+Implementation details
+----------------------
+
+Sequence diagram
+****************
+The communication between SSSD and GDM is explained in the following sequence
+diagram.
+
+.. uml:: ../diagrams/passwordless-gdm-auth.pu
+
+* GDM prompts the user for their username.
+
+* The user initiates the login procedure by entering their username.
+
+* GDM starts the PAM conversation.
+
+* GDM starts the authentication process in SSSD.
+
+* pam_sss communicates with the PAM responder to resolve the user.
+
+* PAM responder resolves the username and obtains, among other things, the
+  available authentication methods and the prompting strings.
+
+* PAM responder prepares the JSON message with the available authentication
+  mechanisms, the prompts and the user credentials to be requested
+  (check format in :ref:`data`).
+
+* PAM responder returns this information to pam_sss.
+
+* pam_sss wraps the JSON message in the PAM conversation using the macros
+  provided by GDM in ``/usr/include/gdm/gdm-custom-json-pam-extension.h``.
+
+* GDM prompts the user for the user credentials, and at the bottom of the
+  screen it shows the available authentication mechanisms. If the user selects
+  another method, GDM already has all the information needed to request the
+  user credentials for this authentication mechanism.
+ 
+* The authentication workflow continues, but varies depending on the selected
+  authentication mechanism. This will be explained separately for each
+  mechanism.
+
+krb5_child
+**********
+``krb5_child`` is the helper binary in charge of Kerberos authentication. It
+follows the general model of separating the authencation in two steps:
+``preauthentication`` and ``authentication``. The first part opens a connection
+to the KDC to receive the Kerberos pre-authentication methods available for the
+given user.
+
+Currently, ``krb5_child`` behaves differently depending if there are
+pre-authentication methods which have to keep the state. Nowadays this would
+be EIdP and passkey. If such a method is found it is immediately selected by
+``krb5_child``, which keeps running and hence keeps the state while the
+information is displayed to the user and they follow the necessary steps for
+authentication. At this point is when the status changes to ``authentication``
+and the still running ``krb5_child`` proceeds with the authentication itself.
+
+If there are no ``stateful`` authentication methods found (password, 2-factor
+authentication, Smartcard/pkinit) a list of available methods are returned by
+``krb5_child`` at the end of ``preauthentication`` and ``krb5_child`` exits.
+The PAM responder selects an authentication method based on some heuristics,
+e.g. if pkinit is available and a Smartcard is present pkinit is selected. Now
+the suitable prompt is displayed to the user and after they entered the
+required credentials the backend starts a new ``krb5_child`` for
+``authentication``. After the KDC has returned the available Kerberos
+pre-authentication method ``krb5_child`` uses the pre-authentication methods
+matching the given credentials to finish the authentication.
+
+This was a valid solution when SSSD was the one deciding which authentication
+method to use during the process. This is no longer the case, since with this
+new proposal it is the user who decides the mechanism to be used, so the
+current ``krb5_child`` design must be extended.
+
+During the ``preauthentication`` phase, ``krb5_child`` opens a connection to
+the KDC to receive the Kerberos pre-authentication methods available for the
+given user and all necessary information (e.g. login URLs, codes, prompts).
+At this point ``krb5_child`` is kept alive for all methods as it must wait for
+the user response. The information is displayed to the user and once they enter
+the credentials, pam_sss switches to the ``authentication`` phase and PAM
+responder serializes the credentials in the ``sss_auth_token`` structure.
+The still running ``krb5_child`` gets the authentication type and the
+credentials, and proceeds with the authentication itself.
+
+.. _data:
+
+Data
+****
+In addition, the messages exchanged follow the JSON standard. SSSD provides the
+available authentication mechanisms in the following message:
+
+..  code-block:: json
+
+    {
+      "authSelection": {
+        "mechanisms": {
+          "$mech1": {
+            "name": "$name1",
+            "role": "$role1",
+            "msg1": "$msg1"
+          },
+          "$mech2": {
+            "name": "$name2",
+            "role": "$role2",
+            "msg1": "$msg2",
+            "msg2": "$msg3"
+          }
+        },
+        "priority": [
+          "$role2",
+          "$role1"
+        ]
+      }
+    }
+
+The field meaning is as follows:
+
+* name: the mechanism name that will be shown in the login screen.
+* role: the authentication mechanism. It can be password, EIdP, smartcard or
+  passkey.
+
+GDM answers with the result for the previous JSON message processing:
+
+..  code-block:: json
+
+    {
+      "authSelection": {
+        "status": "$status",
+        "mech": {
+          "data1": "$data1",
+          "data2": "$data2"
+        }
+      }
+    }
+
+For clarification, there are two different types of user interaction messages:
+`prompt` and `instruction`. The first one asks the user to enter something on
+the screen, e.g. the PIN. The second type prompts the user to follow
+instructions, e.g. to enter the PIN into the reader.
+
+The JSON request and reply messages for different authentication mechanisms are
+defined in the next sections.
+
+Password
+++++++++
+SSSD provides the available authentication mechanisms:
+
+..  code-block:: json
+
+    {
+      "authSelection": {
+        "mechanisms": {
+          "password": {
+            "name": "Password",
+            "role": "password",
+            "prompt": "Password:"
+          }
+        },
+        "priority": [
+          "password"
+        ]
+      }
+    }
+
+GDM replies with the selection:
+
+..  code-block:: json
+
+    {
+      "authSelection": {
+        "status": "Ok",
+        "password": {
+          "password": "ThePassword"
+        }
+      }
+    }
+
+EIdP
+++++
+SSSD provides the available authentication mechanisms:
+
+..  code-block:: json
+
+    {
+      "authSelection": {
+        "mechanisms": {
+          "eidp": {
+            "name": "Web Login",
+            "role": "eidp",
+            "initPrompt": "Login",
+            "linkPrompt": "Log in online with another device",
+            "uri": "https://short.url.com/1234",
+            "code": "1234",
+            "timeout": 300
+          }
+        },
+        "priority": [
+          "eidp"
+        ]
+      }
+    }
+
+GDM replies with the selection:
+
+..  code-block:: json
+
+    {
+      "authSelection": {
+        "status": "Ok",
+        "eidp": {}
+      }
+    }
+
+Smart card
+++++++++++
+SSSD provides the available authentication mechanisms:
+
+..  code-block:: json
+
+    {
+      "authSelection": {
+        "mechanisms": {
+          "smartcard": {
+            "name": "Smartcard",
+            "role": "smartcard",
+            "certificates": [
+              {
+                "tokenName": "sc1",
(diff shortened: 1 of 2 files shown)


src/design-pages/passwordless_gdm.rst
```diff
@@ -0,0 +1,243 @@
...
+++++
+..  code-block:: json
+
+    {
+      "auth-selection": {
+        "mechanisms": [
+          {
+            "eidp": {
+              "init_prompt": "EIdP login",
+              "link_prompt": "EIdP login",
```
ikerexxe, line 190: Improve prompting wording

src/design-pages/passwordless_gdm.rst
```diff
@@ -0,0 +1,367 @@
+Passkey authentication Kerberos integration
```
abbra, line 1: Should the design document title be updated?

ikerexxe: yes, I had done this change locally but I didn't want to push the changes yet. Once I do I'll close this issue.

ikerexxe: Done

src/design-pages/passwordless_gdm.rst
```diff
@@ -0,0 +1,461 @@
...
+  Default is: “PIN”
+
+* pin_instruction=your instruction here: the message that appears to enter the
+  PIN in the reader. Default is: “Enter PIN in reader”
+
+For the passkey mechanism, another section called [prompting/passkey] is
+added. The list of options and their meaning:
+
+* pin_prompt=your prompt here: the message that appears to enter the PIN.
+  Default is: “Security key PIN”
```
justin-stephenson, line 450: Should we use the same defaults prompts as SSSD ?

~~~
src/responder/pam/pam_prompting_config.c:28:#define DEFAULT_PASSKEY_PROMPT_INTERACTIVE _("Insert your Passkey device, then press ENTER.")
src/responder/pam/pam_prompting_config.c:29:#define DEFAULT_PASSKEY_PROMPT_TOUCH _("Please touch the device.")
~~~

ikerexxe: If possible, I would prefer to keep the same configuration options to adjust all user interaction messages. I will discuss with Andre who from the UX team can help us resolve this issue.

justin-stephenson: Okay feel free to 'Resolve' this conversation when you would like to.

src/design-pages/passwordless_gdm.rst
```diff
@@ -0,0 +1,461 @@
...
+SSSD and `GDM <https://wiki.gnome.org/Projects/GDM>`__ are working together to
+provide a set of interfaces that can be used to enable these authentication
+mechanisms in Linux’s GUI. While the initial work targets SSSD-GDM integration,
+the objective is that these interfaces can be used by any other desktop
+environment.
+
+The use of new tools to authenticate users, such as 2FA, U2F and FIDO2, is
+becoming increasingly popular. Currently SSSD provides local authentication of
+a centrally managed user with passkeys, but it doesn't provide any way to
+authenticate remotely to a system. This diminishes the value provided by SSSD
```
sumit-bose, line 34: Hi,

if you want to keep this paragraph (since it is not strictly related to the SSSD GDM integration I think it can be dropped as well) I would like to suggest to mention the FreeIPA Kerberos integration where the Kerberos ticket can be used to connect to remote systems managed by FreeIPA.

bye,
Sumit

src/design-pages/passwordless_gdm.rst
```diff
@@ -0,0 +1,461 @@
...
+        }
+      }
+    }
+
+For clarification, there are two different types of user interaction messages:
+`prompt` and `instruction`. The first one asks the user to enter something on
+the screen, e.g. the PIN. The second type prompts the user to follow
+instructions, e.g. to enter the PIN into the reader.
+
+Examples for the two messages are defined in the following lines.
```
sumit-bose, line 188: Hi,

imo the wording is a bit confusing. In the paragraph before you say "...there are two different types of user interaction messages ..." and to me it sounds like "Examples for the two messages ..." refers to those. Additionally "Examples ... are defined" sounds a bit odd.

Maybe better say something like "Examples for the JSON request and reply messages are given in the following sections" or "The JSON request and reply messages for different authentication mechanisms are defined in the next sections."

bye,
Sumit

src/design-pages/passwordless_gdm.rst
```diff
@@ -0,0 +1,461 @@
...
+            "init_prompt": "Insert smartcard",
+            "pin_prompt": "PIN"
+          },
+          "smartcard:2": {
+            "name": "smartcard ID2",
+            "role": "smartcard",
+            "selectable": true,
+            "init_prompt": "Insert smartcard",
+            "pin_instruction": "Enter PIN in reader"
+          }
```
sumit-bose, line 284: Hi,

I think an "instruction prompt" is missing here, at least if there are multiple matching certificates for the user. It is what is called just `prompt` in SSSD's `pack_cert_data()` and is current a combination of the certificate label and the subject DN of the certificate. To be able to select the right certificate this string should be displayed together with the other details of the given method.

bye,
Sumit

src/design-pages/passwordless_gdm.rst
```diff
@@ -0,0 +1,461 @@
...
+      }
+    }
+
+GDM replies with the selection, as explained in the preceding sections.
+
+Authentication mechanisms priority
+**********************************
+
+The priority for the authentication mechanisms set in the JSON message is a
+hardcoded value and follows the next priority:
```
sumit-bose, line 402: Hi,

why is this not configurable? And how does this relate to the GDM feature to save the last used method. Will the priority only be used if there is no saved method or will the priority always overwrite the last saved method? Or would it make sense to have an option to control this as well?

bye,
Sumit

src/design-pages/passwordless_gdm.rst
ikerexxe: I changed it to your last proposal

src/design-pages/passwordless_gdm.rst
ikerexxe: That's taken care of in the [name](https://github.com/SSSD/sssd.io/pull/79/files#diff-3682413448fe17192d0456526586fa27b75df6ad1dea69372800f974e4203c02R266) attribute. I can change the example to make it more clear, but for that I'll need your help as so far I didn't find any good example.

sumit-bose: Hi,
you can find examples in the PAM responder unit tests (https://github.com/SSSD/sssd/blob/master/src/tests/cmocka/test_pam_srv.c#L87), so e.g.

    "SSSD test cert 0001\nCN=SSSD test cert 0001,OU=SSSD test,O=SSSD"

Please note the `\n`, so on the console it is displayed on two lines, first line the Smartcard label and the second line the subject DN of the certificate. Currently this is hard-coded but it might be worth to make the string configurable. So GDM should not expect a fixed number of lines.

bye,
Sumit

src/design-pages/passwordless_gdm.rst
sumit-bose: Hi,

I wonder if there might be a use-case where an environment might want to emphasize a specific authentication method but allow others as a fall back. So that always the company's preferred method is shown by default independent what the user used last time. It might be possible to implement this with a `forced` flag or similar which is send in `priority` together with the list of methods? But I'm not sure if this is a valid use-case or not.

bye,
Sumit

ikerexxe: As I thought this use case is very hypothetical. That's why I think it's better to leave it hardcoded for the moment, and if in the future we see the need to tune it we can always add a new option to change these values.

I have updated this paragraph to mention the GDM feature to remember the authentication mechanism. Please take a look at it.

src/design-pages/passwordless_gdm.rst
ikerexxe: Actually, this part of the code is already implemented and I am using the [label](https://github.com/SSSD/sssd/blob/master/src/sss_client/pam_sss.c#L135) field of the cert_auth_info structure. Does it contain the label and the subject DN, or just the label?

src/design-pages/passwordless_gdm.rst
sumit-bose: Thanks

src/design-pages/passwordless_gdm.rst
sumit-bose: Hi,

better would be `prompt_str` because `label` would be the same for all certificates from the same Smartcard. `prompt_str` is used in the [current GDM related code](https://github.com/SSSD/sssd/blob/master/src/sss_client/pam_sss.c#L2060) and for[ text based prompting](https://github.com/SSSD/sssd/blob/master/src/sss_client/pam_sss.c#L2152).

bye,
Sumit

ikerexxe: So I'll do that for the implementation. Does this part of the design seem clear and OK to you?

sumit-bose: Hi,

yes, I'm fine, it just might be worth to make the GDM developers aware that the `name` field might have multiple lines.

bye,
Sumit

ikerexxe: @halfline just so you're aware, smartcard's `name` field might have multiple lines.

Sumit, I updated the design to include this information, so I'm resolving the conversation.

sumit-bose:
Hi,

thank you for the updates, I have no further comments.

bye,
Sumit

src/design-pages/passwordless_gdm.rst
ikerexxe: I discussed this topic with Allan and taking into account that the user interaction is different (i.e. passkey CLI expects the user to press enter vs. passkey GUI GDM detects device insertion and changes screen) it's difficult to have common strings for prompts. We even discussed splitting strings, but GNOME style [advises against it](https://developer.gnome.org/hig/guidelines/writing-style.html#translations).

Last Friday Sumit, ab, Andre and I spoke about this topic and we agreed that the best path forward is to concentrate on the passwordless-GDM integration and leave the string unification for a later date. Thus, I'm removing this section and we'll have the fixed strings proposed by Allan in the [os-mockups](https://gitlab.gnome.org/Teams/Design/os-mockups/-/issues/220)

ikerexxe:
I updated the design with new attributes in the smartcard mechanism. They are necessary since the certificate information isn't available during the AUTH phase and we prefer to avoid running `p11_child` again. Thus, we are using the JSON-based communication as a shared memory to hold the certificate data that is needed later on.

ikerexxe:
Updated design with additional attributes needed for passkey authentication.

src/design-pages/passwordless_gdm.rst
```diff
@@ -117,6 +117,32 @@ diagram.
   authentication mechanism. This will be explained separately for each
   mechanism.
 
+krb5_child
+****************
```
pbrezina, line 121: to many asterisks

src/design-pages/passwordless_gdm.rst
```diff
@@ -117,6 +117,32 @@ diagram.
...
+
+During the ``preauthentication`` phase all authentication methods that are
+available to the user are checked and all necessary information (e.g. login
+URLs, codes, prompts) is obtained to proceed with the authentication.
+Since some authentication mechanisms need to keep a session open, these are
+marked as keep-alive. The information is displayed to the user and once
+the user enters the credentials, pam_sss switches to the ``authentication``
+phase and PAM responder serializes the credentials in the ``sss_auth_token``
+structure. ``krb5_child`` gets the authentication type and the credentials,
+and continues with the authentication process.
```
pbrezina, line 144: Are you describing the current state or the future state here?

ikerexxe: Future state. Am I missing anything?

pbrezina: "Since some authentication mechanisms need to keep a session open, these are
marked as keep-alive." Keep krb5_child alive for all methods to unify the behavior.

This is the sentence that confused me, since this is the current behavior - only selected methods keep krb5_child running. Now we want the "preauthentication" to obtain the list and all information about all methods, keep krb5_child alive and wait for response.

We could also rename "preauthentication" to "get prompts" or "list methods" to make its meaning more  clear.

src/design-pages/passwordless_gdm.rst
ikerexxe: Fixed

src/design-pages/passwordless_gdm.rst
ikerexxe: I've rewritten the sentence to make it clear that `krb5_child` is kept alive for all methods.

> We could also rename "preauthentication" to "get prompts" or "list methods" to make its meaning more clear.

I would prefer to avoid this change since this reference is used in more than ten files and I am making these changes outside of the main SSSD development branch, and will probably have to continue this way for a few more months. I've been working this way for about a year and a half now and it's bearable because my changes are fairly self-contained, but I've had a scare or two, and making this change would break that dynamic.

src/design-pages/passwordless_gdm.rst
```diff
@@ -117,6 +117,32 @@ diagram.
...
+krb5_child
+**********
+``krb5_child`` is the helper binary in charge of Kerberos authentication. It
+follows the general model of separating the authencation in two steps:
+``preauthentication`` and ``authentication``. The first part opens a session
+to obtain the data related to the authentication (i.e. EIdP code). The
+``krb5_child`` keeps the state while the information is displayed to the user
+and they follow the necessary steps for authentication. At this point is when
+the status changes to ``authentication`` and proceeds with the authentication
+itself.
```
sumit-bose, line 129: Hi,

if I understand it correctly the first paragraph should describe the current state. This is imo incomplete and as a result I think it is not completely clear why and what a change is needed.

I would like to suggest to following changes:

.... The first part opens a connection to the KDC to receive the Kerberos pre-authentication methods available for the given user. ``krb5_child`` will behave differently depending if there are pre-authentication methods which have to keep the state (currently EIdP and passkey). If such method is found it is immediately selected by ``krb5_child`` which keeps running and hence keeps the state while the information is displayed to the user
and they follow the necessary steps for authentication. At this point is when the status changes to ``authentication`` the still running ``krb5_child`` proceeds with the authentication itself.

If there are no "stateful" authentication methods found (password, 2-factor authentication, Smartcard/pkinit) a list of available methods are return by ``krb5_child`` at the end of ``preauthentication` and ``krb5_child`` exits. The PAM responder selects an authentication method based on some heuristics, e.g. is pkinit is available and a Smartcard is present pkinit is selected. Now a suitable prompt is displayed to the user and after they entered the required credentials the backend starts a new ``krb5_child`` now for ``authentication``. After the KDC has returned the available Kerberos pre-authentication method ``krb5_child` uses the pre-authentication methods matching the given credentials to finish the authentication.


HTH

bye,
Sumit

ikerexxe: Thank you very much for the explanation, this helps to understand better how everything works nowadays. I have added this information, with slight modifications, to the text.

ikerexxe:
@pbrezina @sumit-bose I updated the text with the improvements proposed. Take another look to make sure everything is correct.

sumit-bose:
> @pbrezina @sumit-bose I updated the text with the improvements proposed. Take another look to make sure everything is correct.

Hi,

thanks for the updates, the "krb5_child" section is looking good, ACK.

bye,
Sumit

ikerexxe:
As discussed with Joan I'm updating the design with the following changes:
* Update attributes to camel case
* Remove `initInstruction` for smartcard as this won't be needed
* Convert the smartcard list to an actual JSON array

src/design-pages/passwordless_gdm.rst
```diff
@@ -0,0 +1,489 @@
...
+              "certInstruction": "Certificate for PIV Authentication\nCN=sc2,O=GDM.TEST",
+              "pinInstruction": "Enter PIN in reader",
+              "moduleName": "/usr/lib64/pkcs11/opensc-pkcs11.so",
+              "keyId": "02",
+              "label": "Certificate for PIV Authentication"
+            }
+          ]
+        },
+        "priority": [
+          "smartcard:1",
```
joantolo, line 331: I think here it should only be the mechanism key, i.e., "smartcard".

src/design-pages/passwordless_gdm.rst
```diff
@@ -0,0 +1,489 @@
...
+    }
+
+All authentication mechanisms
++++++++++++++++++++++++++++++
+SSSD provides the available authentication mechanisms:
+
+..  code-block:: json
+
+    {
+      "authSelection": {
```
joantolo, line 406: The others had auth-selection. All could be authSelection, or is this a special keyword?

src/design-pages/passwordless_gdm.rst
```diff
@@ -0,0 +1,489 @@
...
+            "pinRequest": true,
+            "pinAttempts": 8,
+            "pinPrompt": "Security key PIN",
+            "touchInstruction": "Touch security key",
+            "kerberos": true,
+            "cryptoChallenge": "6uDMvRKj3W5xJV3HaQjZrtXMNmUUAjRGklFG2MIhN5s="
+          }
+        },
+        "priority": [
+          "smartcard:1",
```
joantolo, line 450: Now I think this should be smartcard instead.

src/design-pages/passwordless_gdm.rst
```diff
@@ -0,0 +1,489 @@
...
+              "pinPrompt": "PIN",
+              "moduleName": "/usr/lib64/pkcs11/opensc-pkcs11.so",
+              "keyId": "01",
+              "label": "Certificate for PIV Authentication"
+            }
+          ],
+          "passkey": {
+            "name": "Passkey",
+            "role": "passkey",
+            "selectable": true,
```
joantolo, line 439: One thing that was [discussed](https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/8513#note_2539629) is making selectable property implicit, so if you want you can drop this property.

We'll just assume what's selectable based on the role:
* passkey: selectable
* smartcard: selectable
* eidp: selectable
* password: selectable
* fingerprint: non selectable
* faceunlock: non selectable

src/design-pages/passwordless_gdm.rst
```diff
@@ -0,0 +1,489 @@
...
+..  code-block:: json
+
+    {
+      "auth-selection": {
+        "mechanisms": {
+          "passkey": {
+            "name": "Passkey",
+            "role": "passkey",
+            "selectable": true,
+            "initInstruction": "Insert security key",
```
joantolo, line 369: This won't be used until passkey detection is implemented, you can drop it if you want.

src/design-pages/passwordless_gdm.rst
```diff
@@ -0,0 +1,489 @@
...
+
+    {
+      "auth-selection": {
+        "status": "Ok",
+        "smartcard:1": {
+          "pin": "ThePIN",
+          "name": "sc1",
+          "moduleName": "/usr/lib64/pkcs11/opensc-pkcs11.so",
+          "keyId": "01",
+          "label": "Certificate for PIV Authentication"
```
joantolo, line 351: Just a question, not necessarly a need to change:

Are all these properties needed to verify we are sending the pin for the correct certificate? Maybe `smartcard:1` with `pin` is enough? What about "name" and "label"?

src/design-pages/passwordless_gdm.rst
ikerexxe: Yes, that's right

src/design-pages/passwordless_gdm.rst
ikerexxe: Ok, that sounds good to me. I had to hardcode it anyway

src/design-pages/passwordless_gdm.rst
ikerexxe: Did you discuss this topic with Allan?

src/design-pages/passwordless_gdm.rst
```diff
@@ -0,0 +1,482 @@
...
+
+    {
+      "authSelection": {
+        "mechanisms": {
+          "smartcard": {
+            "name": "Smartcard",
+            "role": "smartcard",
+            "certificates": [
+              {
+                "id": "smartcard:1",
```
joantolo, line 303: I know I asked this id property, but after thinking a bit, let's just drop "id" here if we're not going to use it.

src/design-pages/passwordless_gdm.rst
```diff
@@ -0,0 +1,479 @@
...
+login as there are no saved methods. Once the user authenticates with a method
+GDM will remember it and the next time the same user tries to login this method
+will be provided. The priority is as follows:
+
+* Smartcard
+* Passkey
+* EIdP
+* Password
+
+Configuration options
```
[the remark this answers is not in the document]
ikerexxe: Done

ikerexxe:
> @ikerexxe once released will Passwordless GDM supported package versions be added to to this design page? (SSSD, GDM, authselect)

@justin-stephenson it makes sense so I added those at the end

justin-stephenson (dismissed):

src/design-pages/passwordless_gdm.rst
```diff
@@ -0,0 +1,504 @@
...
+
+* As a centrally managed user, I want to select the smart card identity to
+  login from the graphical interface so that the authentication is performed
+  with the correct credentials.
+
+* As a centrally managed user, I want to use a passkey to login from the
+  graphical interface so that the user interface is easy to use and consistent
+  across all authentication mechanisms in the distribution.
+
+* As a centrally managed user, I want to get notified when the passkey
```
sumit-bose: Hi,

I think the same is true for the fallback to local Smartcard authentication. Not sure if this is already handled by the current code. If not this might be something to be added in future.

bye,
Sumit

ikerexxe: Yes, pam_sss informs about this for passkey and it's shown by GDM after authentication and before moving to the "home" screen. If smartcard does the same, then we are covered, if not we should implement it.

In any case, I've added this to the design.

sumit-bose (approved):
Hi,

thank you for the updates, ACK.

bye,
Sumit

PR #80, opened 2024-02-06
Design pages: fix userId in passkey

ikerexxe:
userId is a private attribute and it shouldn't be stored in a public LDAP attribute. Thus, it was decided to stop printing it and storing it in LDAP, but I forgot to update the design. This commit fixes this issue.

Resolves: https://github.com/SSSD/sssd/issues/7165


justin-stephenson (approved):
Ack, thank you.

PR #91, opened 2025-02-18
Adding a profiling page

aplopez:
Adding a page describing how to profile any of the SSSD daemons.

The page be seen here: https://sssdio--91.org.readthedocs.build/contrib/profiling-sssd.html

### DIFF
--- a/src/contrib/profiling-sssd.rst
+++ b/src/contrib/profiling-sssd.rst
@@ -1,0 +1,113 @@
+Profiling SSSD
+##############
+
+Getting Ready
+*************
+
+There are several tools allowing us to profile an application. We will focus on
+perf_ which seems to be the one that works best with SSSD.
+
+.. _perf: https://perf.wiki.kernel.org
+
+First you need to install the ``perf`` tool:
+
+.. code-tabs::
+
+    .. fedora-tab::
+
+        # dnf -y install perf js-d3-flame-graph
+
+    .. ubuntu-tab::
+
+        $ sudo apt install linux-tools-common linux-tools-generic
+
+
+Make sure the debug information is available for SSSD and all its dependencies
+either by using ``debuginfod``, or by manually installing the debug information
+packages:
+
+.. code-tabs::
+
+    .. fedora-tab::
+
+        # dnf debuginfo-install -y sssd* libsss_* samba-client-libs libsmbclient \
+        libldb libtevent libtalloc libini_config c-ares glibc
+
+    .. ubuntu-tab::
+
+        $ sudo apt install sssd-ad-common-dbgsym sssd-ad-dbgsym sssd-common-dbgsym \
+        sssd-dbus-dbgsym sssd-ipa-dbgsym sssd-kcm-dbgsym sssd-krb5-common-dbgsym \
+        sssd-krb5-dbgsym sssd-ldap-dbgsym sssd-proxy-dbgsym sssd-tools-dbgsym \
+        libsss-certmap0-dbgsym libsss-idmap0-dbgsym libsss-nss-idmap0-dbgsym \
+        libsss-simpleifp0-dbgsym libsss-sudo-dbgsym libldb2-dbgsym libtdb1-dbgsym \
+        libtalloc2-dbgsym libtevent0-dbgsym libc6-dbgsym
+
+.. seealso::
+
+    More information on `Debug Symbol Packages`_ for Ubuntu.
+
+.. _`Debug Symbol Packages`: https://documentation.ubuntu.com/server/reference/debugging/debug-symbol-packages/
+
+.. note::
+
+    If you are using SELinux, the shipped targeted policy may prevent ``perf``
+    from interacting with the SSSD process. You can put SELinux into permissive
+    mode to confirm or look at the audit logs and add the required rules. Rules
+    can be added using ``audit2allow``. 
+
+Profiling
+*********
+
+The simplest way to profile one of the SSSD's daemons is to attach the profiler
+to the process while it is running.
+
+Once SSSD is running and ready to be profiled, identify the PID of the process
+you want to monitor (``sssd_ssh``, ``sssd_nss``, ``sssd_be``, etc.), start the
+``perf`` command in the background, launch the command you want to profile
+and stop the profiling after the command completes:
+
+.. code-block:: console
+    :caption: `Profiling the LDAP domain daemon during the id command`
+
+    # PID=$(pgrep -f 'sssd_be --domain LDAP')
+    # perf record --pid=$PID --call-graph=dwarf -e cycles:u &
+    # id user1002@LDAP
+    # kill %%
+
+.. code-block:: console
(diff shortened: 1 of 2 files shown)


src/contrib/profiling-sssd.rst
```diff
@@ -0,0 +1,108 @@
...
+
+.. seealso::
+
+    More information on `Debug Symbol Packages`_ for Ubuntu.
+
+.. _`Debug Symbol Packages`: https://documentation.ubuntu.com/server/reference/debugging/debug-symbol-packages/
+
+.. note::
+
+    It is possible that SELinux prevents ``perf`` from monitoring your process.
```
danlavu, line 53: Is this certain? That SELinux rules block perf from interacting with the SSSD domain?  If it's for certain, then we should phrase it that way.

`The shipped targetted SELinux policy will prevent ``perf`` from interacting with the SSSD process. You can look at the audit logs or put SELinux into permissive mode to confirm. You will need to add additional rules and can be done using ``audit2allow``. `

src/contrib/profiling-sssd.rst
```diff
@@ -0,0 +1,108 @@
...
+
+
+Profiling
+*********
+
+The simplest way to profile one of the SSSD's daemons is to attach the profiler
+to the process while it is running.
+
+Once SSSD is running and ready to be profiled, identify the PID of the process
+you want to monitor (``sssd_ssh``, ``sssd_nss``, ``sssd_be``, etc.), start the
```
danlavu, line 69: It would be nice to provide more examples, as in what data will be profiled selected the common processes. We will make the assumption that administrators may fully understand the SSSD architecture.

danlavu (changes requested):
Other than the few comments, it looks good.

src/contrib/profiling-sssd.rst
```diff
@@ -0,0 +1,108 @@
...
+    .. fedora-tab::
+
+        # dnf -y install perf js-d3-flame-graph
+
+    .. ubuntu-tab::
+
+        $ sudo apt install linux-tools-common linux-tools-generic
+
+
+Make sure SSSD's and its dependencies' debug information is available, either
```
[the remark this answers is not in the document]
aplopez: Fixed (I hope).

src/contrib/profiling-sssd.rst
aplopez: I slightly modified it as this text is meant for any distribution, not only RHEL or Fedora (although the given examples are for Fedora and Ubuntu). How does it look?

src/contrib/profiling-sssd.rst
aplopez: Added one more example.

aplopez:
Fixed the comments and also some small errors I found.

src/contrib/profiling-sssd.rst
```diff
@@ -0,0 +1,116 @@
...
+    More information on `Debug Symbol Packages`_ for Ubuntu.
+
+.. _`Debug Symbol Packages`: https://documentation.ubuntu.com/server/reference/debugging/debug-symbol-packages/
+
+.. note::
+
+    If you are using SELinux, the shipped targeted policy may prevent ``perf``
+    from interacting with the SSSD process. You can put SELinux into permissive
+    mode to confirm or look at the audit logs and add the required rules. Rules
+    can be added using ``audit2allow``.
```
danlavu, line 57: Technically, audit2allow generates a module that can be loaded. That detail is not important.

danlavu (approved):
One comment, about the SELinux language, if someone needs help with it, there is always a web search, so it's not important. Thank you Alejandro

src/contrib/profiling-sssd.rst
aplopez: I think we can trust that users adding SELinux rules know what the do. Non-expert users will probably set it to permissive mode while profiling.

src/contrib/profiling-sssd.rst
```diff
@@ -0,0 +1,116 @@
...
+either by using ``debuginfod``, or by manually installing the debug information
+packages:
+
+.. code-tabs::
+
+    .. fedora-tab::
+
+        # dnf -y --enable-repo=fedora-debuginfo --enable-repo=updates-debuginfo \
+        install sssd*debuginfo libsss_*-debuginfo samba-client-libs-debuginfo \
+        libldb-debuginfo libtevent-debuginfo libtalloc-debuginfo glibc-debuginfo
```
alexey-tikhonov, line 35: Maybe `dnf debuginfo-install sssd`?

alexey-tikhonov: Or will it miss debug info for deps..?

alexey-tikhonov: Also `libcares`, `libtdb`, `libdhash`, `libldap`...

src/contrib/profiling-sssd.rst
```diff
@@ -0,0 +1,116 @@
...
+Generating the Reports
+**********************
+
+We will create two types of reports: a text report and a flame graph to be seen
+in a web browser. But before doing that, it is necessary to update ``perf``'s
+cache of debug information:
+
+.. code-block:: console
+
+    # perf buildid-list
```
alexey-tikhonov, line 105:
```
This command displays the buildids found in a perf.data file, so that other tools can be used to fetch packages with matching symbol tables for use by perf report.
```

It doesn't say anything about cache update...

src/contrib/profiling-sssd.rst
aplopez: I didn't know the first command. I think it is better to use it instead of repeating `-debuginfo` on every package. Nevertheless, it also install the `-debugsource` packages which, I think, are not required. I will check that.

In addition, it doesn't resolve the dependencies, so every single package must still be mentioned explicitly.

Debuginfo for `libtdb` is included in `libldb-debuginfo`. Iḿ adding the other you mentioned and a few more.

aplopez: After checking, `debuginfo-install` does resolve the dependencies for other packages. So I guess there might be some problem with ours.

src/contrib/profiling-sssd.rst
aplopez: You are right. I read that somewhere and, at that moment, it seemed to solve my problem. However, now I can no longer reproduce that problem and this command seems to be useless.

src/contrib/profiling-sssd.rst
alexey-tikhonov: > sssd* libsss_*

Doesn't simple "sssd" work (installing everything, at least for SSSD)?

alexey-tikhonov: Looks like "no".

alexey-tikhonov (approved):

aplopez:
@danlavu It seems the configuration of this repository requires that you approve again this PR because I applied some changes. Sorry for this.

